Issue #8, opened 2017-04-07
Test is weird


diasdavid:
@dignifiedquire this test

```
  1) DHT find peer query:

      Uncaught AssertionError: expected 1 to deeply equal 0
      + expected - actual

      -1
      +0

      at utils.sortClosestPeers (test/index.spec.js:240:51)
      at map (src/utils.js:120:5)
      at node_modules/async/internal/map.js:32:9
      at node_modules/async/internal/once.js:12:16
      at iteratorCallback (node_modules/async/eachOf.js:60:13)
      at node_modules/async/internal/onlyOnce.js:12:16
      at node_modules/async/internal/map.js:29:13
      at exports.convertPeerId (src/utils.js:110:7)
      at setImmediate (node_modules/multihashing-async/src/utils.js:8:7)
      at Immediate.<anonymous> (node_modules/async/internal/setImmediate.js:26:16)

```

sometimes fail. If we look at the test, we can see that it sets up a random DHT leading to sometimes non predictable scenarios.

diasdavid:
fixed in https://github.com/libp2p/js-libp2p-kad-dht/pull/25

PR #13, opened 2017-07-16
fix: no more circular dependency, become a good block of libp2p

### LINKED ISSUE #9: Circular dep with libp2p-ipfs-nodejs
This module is a piece of js-libp2p bundles, but at the same time it expects to receive a bundle to operate, forcing us to have to create an instance inside the base class. This is annoying to work with because it forces updating and development to happen through `npm link` with a bundle for changes that require modifying how the bundle gets created. This module should use the internal components and not the bundle itself.

diasdavid:
Fix #9. Just one last thing, fix:

```
 1) KadDHT put - get:
   Uncaught TypeError: pi.id.toB58String is not a function
    at Swarm.dial (node_modules/libp2p-swarm/src/dial.js:22:25)
    at Network.sendRequest (src/network.js:143:20)
    at KadDHT._findPeerSingle (src/private.js:244:17)
```

by leveling up swarm.dial to also support PeerId instead of just PeerInfo.


src/index.js
```diff
@@ -32,66 +33,71 @@ class KadDHT {
...
      * k-bucket size, defaults to 20.
      *
      * @type {number}
      */
-    this.kBucketSize = kBucketSize || 20
+    this.kBucketSize = options.kBucketSize || 20
 
     /**
-     * Number of closest peers to return on kBucket search
+     * Number of closest peers to return on kBucket search, default 20
```
dignifiedquire: this is not 20 in the code but 6

test/kad-dht.spec.js
```diff
@@ -11,23 +11,26 @@ const timeout = require('async/timeout')
 const retry = require('async/retry')
```
dignifiedquire: 👎 for the rename

src/index.js
```diff
@@ -28,70 +29,74 @@ class KadDHT {
...
    *
-   * @param {Libp2p} libp2p
-   * @param {number} [kBucketSize=20]
-   * @param {Datastore} [datastore=MemoryDatastore]
+   * @param {swarm} Swarm
+   * @param {options} {kBucketSize=20, datastore=MemoryDatastore}
    */
-  constructor (libp2p, kBucketSize, datastore) {
+  constructor (swarm, options) {
+    assert(swarm, 'libp2p-kad-dht requires a instance of kad-dht')
```
pgte: don't you mean "libp2p-kad-dht requires a instance of **swarm**"?

diasdavid:
CI is failing due to previously reported bug: https://github.com/libp2p/js-libp2p-kad-dht/issues/8

PR #33, opened 2018-08-24
Make logs better for users and fix flaky test

jacobheun:
## Log update
The logs are currently returning utf-8 strings for many of the key references in the dht. This makes the data very unhelpful for users. This PR adds a base58 formatter for `debug` via `multihashes` and updates the logs across the file where needed.

## Flaky Test Fix
I also noticed the expire test for providers was flaky because of the timeouts. The interval was set to 100 and the expire was set to 200. The test check was only at 300, which it is possible for the cleanup to trigger after. I've upped the test to 400ms, although being able to monitor the cleanup would be ideal, to prevent the flakiness. I've re-run the tests a couple time to verify, https://ci.ipfs.team/blue/organizations/jenkins/libp2p%2Fjs-libp2p-kad-dht/activity?branch=fix%2Flogs.

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,6 +46,7 @@
     "k-bucket": "^4.0.1",
     "libp2p-crypto": "~0.13.0",
     "libp2p-record": "~0.5.1",
+    "multihashes": "~0.4.14",
     "multihashing-async": "~0.5.1",
     "peer-id": "~0.11.0",
     "peer-info": "~0.14.1",
@@ -57,7 +58,7 @@
     "xor-distance": "^1.0.0"
   },
   "devDependencies": {
-    "aegir": "^15.0.0",
+    "aegir": "^15.1.0",
     "chai": "^4.1.2",
     "datastore-level": "~0.8.0",
     "dirty-chai": "^2.0.1",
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -46,6 +46,7 @@
     "k-bucket": "^4.0.1",
     "libp2p-crypto": "~0.13.0",
     "libp2p-record": "~0.5.1",
+    "multihashes": "~0.4.14",
```
diasdavid: can we use the CID package instead? (Trying to get into the habit to make migration easier)

jacobheun: I think the dht should get updated to use cids across the board since it's just using buffers right now. That will be a pretty decent lift to the code base to make sure we have backwards compatibility. We can either replace this PR with one for that, or follow this PR up with the switch to CID. With the short quarter and higher priorities I think a new issue that we can slate for October might be best. I can create the issue if that sounds good, it should be a good starter ticket for someone if they're able to get to it sooner.

diasdavid: Sounds good!

jacobheun: Created https://github.com/libp2p/js-libp2p-kad-dht/issues/34 for the migration to cids.

vasco-santos: Thanks @jacobheun

PR #39, opened 2018-09-19
Do parallel lookups on disjoint sets of nodes

### LINKED ISSUE #37: Correctly implement SKademlia
This is the js version of https://github.com/libp2p/go-libp2p-kad-dht/issues/146

jhiesey:
This implements one of the core extensions of S/Kademlia.
Fixes #37. Existing tests are passing but new tests are still needed.

Could I get some feedback on this before doing the go version?

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -31,4 +31,8 @@ exports.K = 20
 // Alpha is the concurrency for asynchronous requests
 exports.ALPHA = 3
 
+// Number of disjoint query paths to use
+// This is set to K/2 per the S/Kademlia paper
+exports.DISJOINT_PATHS = 10
+
 exports.maxMessageSize = 2 << 22 // 4MB
--- a/src/index.js
+++ b/src/index.js
@@ -253,7 +253,7 @@ class KadDHT {
     }
 
     this._log('getMany %b (%s)', key, nvals)
-    const vals = []
+    let vals = []
 
     this._getLocal(key, (err, localRec) => {
       if (err && nvals === 0) {
@@ -271,6 +271,7 @@ class KadDHT {
         return callback(null, vals)
       }
 
+      const paths = []
       waterfall([
         (cb) => utils.convertBuffer(key, cb),
         (id, cb) => {
@@ -284,38 +285,49 @@ class KadDHT {
             return cb(errcode(new Error(errMsg), 'ERR_NO_PEERS_IN_ROUTING_TABLE'))
           }
 
-          // we have peers, lets do the actualy query to them
-          const query = new Query(this, key, (peer, cb) => {
-            this._getValueOrPeers(peer, key, (err, rec, peers) => {
-              if (err) {
-                // If we have an invalid record we just want to continue and fetch a new one.
-                if (!(err.code === 'ERR_INVALID_RECORD')) {
-                  return cb(err)
+          // we have peers, lets do the actual query to them
+          const query = new Query(this, key, (pathIndex, numPaths) => {
+            // This function body runs once per disjoint path
+            const pathSize = utils.pathSize(nvals - vals.length, numPaths)
+            const pathVals = []
+            paths.push(pathVals)
+
+            // Here we return the query function to use on this particular disjoint path
+            return (peer, cb) => {
+              this._getValueOrPeers(peer, key, (err, rec, peers) => {
+                if (err) {
+                  // If we have an invalid record we just want to continue and fetch a new one.
+                  if (!(err.code === 'ERR_INVALID_RECORD')) {
+                    return cb(err)
+                  }
                 }
-              }
 
-              const res = { closerPeers: peers }
+                const res = { closerPeers: peers }
 
-              if ((rec && rec.value) || (err && err.code === 'ERR_INVALID_RECORD')) {
-                vals.push({
-                  val: rec && rec.value,
-                  from: peer
-                })
-              }
+                if ((rec && rec.value) || (err && err.code === 'ERR_INVALID_RECORD')) {
+                  pathVals.push({
+                    val: rec && rec.value,
+                    from: peer
+                  })
+                }
 
-              // enough is enough
-              if (vals.length >= nvals) {
-                res.success = true
-              }
+                // enough is enough
+                if (pathVals.length >= pathSize) {
+                  res.success = true
+                }
 
-              cb(null, res)
-            })
+                cb(null, res)
+              })
+            }
           })
 
           // run our query
           timeout((cb) => query.run(rtp, cb), options.maxTimeout)(cb)
         }
       ], (err) => {
+        // combine vals from each path
+        vals = [].concat.apply(vals, paths).slice(0, nvals)
+
         if (err && vals.length === 0) {
           return callback(err)
         }
@@ -341,15 +353,20 @@ class KadDHT {
 
       const tablePeers = this.routingTable.closestPeers(id, c.ALPHA)
 
-      const q = new Query(this, key, (peer, callback) => {
-        waterfall([
-          (cb) => this._closerPeersSingle(key, peer, cb),
-          (closer, cb) => {
-            cb(null, {
-              closerPeers: closer
-            })
-          }
-        ], callback)
+      const q = new Query(this, key, () => {
+        // There is no distinction between the disjoint paths,
+        // so there are no per-path variables in this scope.
+        // Just return the actual query function.
+        return (peer, callback) => {
+          waterfall([
+            (cb) => this._closerPeersSingle(key, peer, cb),
+            (closer, cb) => {
+              cb(null, {
+                closerPeers: closer
+              })
+            }
+          ], callback)
+        }
       })
 
       q.run(tablePeers, (err, res) => {
@@ -549,25 +566,30 @@ class KadDHT {
           }
 
           // query the network
-          const query = new Query(this, id.id, (peer, cb) => {
-            waterfall([
-              (cb) => this._findPeerSingle(peer, id, cb),
-              (msg, cb) => {
-                const match = msg.closerPeers.find((p) => p.id.isEqual(id))
+          const query = new Query(this, id.id, () => {
+            // There is no distinction between the disjoint paths,
+            // so there are no per-path variables in this scope.
+            // Just return the actual query function.
+            return (peer, cb) => {
+              waterfall([
+                (cb) => this._findPeerSingle(peer, id, cb),
+                (msg, cb) => {
+                  const match = msg.closerPeers.find((p) => p.id.isEqual(id))
 
-                // found it
-                if (match) {
-                  return cb(null, {
-                    peer: match,
-                    success: true
+                  // found it
+                  if (match) {
+                    return cb(null, {
+                      peer: match,
+                      success: true
+                    })
+                  }
+
+                  cb(null, {
+                    closerPeers: msg.closerPeers
                   })
                 }
-
-                cb(null, {
-                  closerPeers: msg.closerPeers
-                })
-              }
-            ], cb)
+              ], cb)
+            }
           })
 
           timeout((cb) => {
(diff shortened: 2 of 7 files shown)


jhiesey:
@jacobheun @diasdavid @vasco-santos Any feedback?

src/constants.js
```diff
@@ -31,4 +31,7 @@ exports.K = 20
 // Alpha is the concurrency for asynchronous requests
 exports.ALPHA = 3
 
+// Number of disjoint paths to search down
+exports.DISJOINT_PATHS = 6
```
vasco-santos: Shouldn't this be 16? 

According to [go-libp2p-kad-dht#146#issue-320532077](https://github.com/libp2p/go-libp2p-kad-dht/issues/146#issue-320532077)

> we actually need to query with 16 different paths

src/query.js
```diff
@@ -18,13 +19,15 @@ class Query {
    *
    * @param {DHT} dht - DHT instance
    * @param {Buffer} key
-   * @param {function(PeerId, function(Error, Object))} query - The query function to exectue
+   * @param {function(PeerId, Object, function(Error, Object))} query - The query function to exectue
```
vasco-santos: Please add an explanation for the Object state

Typo: `exectue`

src/query.js
```diff
@@ -18,13 +19,15 @@ class Query {
...
+   * @param {function(Array<Object>)} aggregate - The aggregation function to combine multiple buckets
```
vasco-santos: As far as I understood, this should be `@param {function(Object, function(Error, Array<Object>))}`, right?

src/constants.js
jhiesey: I'm curious where the 16 came from in the original issue.

I pulled the number 6 from the S/Kademlia paper, but I agree it might be a bit low. The actual simulations from the paper don't seem very thorough to be honest, but the authors do suggest d [number of disjoint paths] = k/2: "We conclude that d = 4..8 with a k = 8..16 is a good choice for S/Kademlia". Given that we're using k = 20, maybe we should start with d = 10?

I think we'll have to do simulation and/or testing to know what the right choice actually is.

src/query.js
jhiesey: Sure

src/query.js
jhiesey: Good catch! I think it actually ought to be just `@param {function(Object, function(Error, Object))}` since the result isn't necessarily an array; it's up to the caller to ensure that whatever the aggregate gives to its callback is what `query.run`'s callback expects.

src/constants.js
vasco-santos: yes! It would be nice to make some simulation tests, in order to determine the best number here. By the end of this quarter, we will hopefully have a testbed for large scale tests to the DHT. Maybe there we can have better conclusions about this.

src/query.js
vasco-santos: yes, I agree! 👍 

Change to that then:

```js
@param {function(Object, function(Error, Object))}
```

src/constants.js
vasco-santos: I will try to read the S/Kademlia paper again during the next week. I read it a long time ago, and I hope to be able to provide you better feedback once you add the remaining tests.

jhiesey:
I ended up rewriting this from scratch, but I may not have addressed all issues. At least there is a test this time :)

src/constants.js
jhiesey: New version uses d=10. Thoughts?

src/query.js
jhiesey: New version has better types.

src/query.js
jhiesey: New version should have this fixed.

jhiesey:
All prior comments should be addressed. Can this get another look?

src/constants.js
vasco-santos: I would like to have tests using several different `d`, in order to get the best option. Maybe we should discuss this in the original issue and grab opinion from other folks

src/constants.js
```diff
@@ -31,4 +31,7 @@ exports.K = 20
 // Alpha is the concurrency for asynchronous requests
 exports.ALPHA = 3
 
+// Number of disjoint query paths to use
+exports.DISJOINT_PATHS = 10
```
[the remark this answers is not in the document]
jhiesey: That isn't so easy. The "best" option really depends on network conditions; i.e. if there is someone trying to attack the network.

If every queried peer is answering honestly, then 1 (the original behavior) is best. The [S/Kademlia paper](http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.68.4986&rep=rep1&type=pdf) suggests K/2, which is why I chose 10. However, network traffic will go up proportionally as we increase this constant.

How about we set it to something very low, like 2 or 3, for now? Or, if you don't like that idea, what kind of tests do you have in mind to choose a better value?

src/index.js
```diff
@@ -259,6 +259,9 @@ class KadDHT {
         return callback(null, vals)
       }
 
+      const perPath = Math.ceil((vals.length - nvals) / c.DISJOINT_PATHS)
```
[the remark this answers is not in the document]
jhiesey: Oops that's definitely a bug, thanks!

src/index.js
```diff
@@ -270,39 +273,51 @@ class KadDHT {
...
-          const query = new Query(this, key, (peer, cb) => {
-            this._getValueOrPeers(peer, key, (err, rec, peers) => {
-              if (err) {
-                // If we have an invalid record we just want to continue and fetch a new one.
-                if (!(err instanceof errors.InvalidRecordError)) {
-                  return cb(err)
+          // we have peers, lets do the actual query to them
+          const query = new Query(this, key, () => {
+            const pathVals = []
+            paths.push(pathVals)
```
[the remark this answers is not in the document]
jhiesey: Yes, it's initially empty. Contents are pushed into pathVals inside the following function. Should I clarify this?

src/index.js
```diff
@@ -328,18 +343,20 @@ class KadDHT {
...
+              cb(null, {
+                closerPeers: closer
+              })
+            }
+          ], callback)
+        }
       })
 
-      q.run(tablePeers, (err, res) => {
+      q.run(tablePeers, c.DISJOINT_PATHS, (err, res) => {
```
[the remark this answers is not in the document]
jhiesey: The only case I use a different value is in the tests, but the call to Math.min() takes care of that, so I'll remove this. Good catch!

src/index.js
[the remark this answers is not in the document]
jhiesey: That was supposed to be nvals - vals.length

src/constants.js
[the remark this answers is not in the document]
vasco-santos: Ok, considering that the paper suggests `K/2`, I think we should go with it for now and after this is finished, try to have test scenarios, in order to understand if we should tweak it or not. This way, I only ask you to add a comment here mentioning that we are doing `K/2` as suggested by the paper. Seems good to you?

src/index.js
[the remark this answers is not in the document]
vasco-santos: Ok, I understand what you are doing. So, you add the reference and then push to it. Why not pushing it to the `paths` by the end? I think the logic would be more understandable.

src/index.js
```diff
@@ -273,38 +274,51 @@ class KadDHT {
...
           // run our query
           timeout((cb) => query.run(rtp, cb), options.maxTimeout)(cb)
         }
       ], (err) => {
+        // combine vals from each path
+        paths.forEach((path) => {
+          path.forEach((val) => {
+            vals.push(val)
+          })
+        })
```
vasco-santos: You can use the following here:

```js
vals = vals.concat(...paths)
```

or

```js
vals = [].concat.apply(vals, newVals)
```

src/private.js
```diff
@@ -489,33 +489,47 @@ module.exports = (dht) => ({
...
+        }
       })
 
       const peers = dht.routingTable.closestPeers(key.buffer, c.ALPHA)
 
       timeout((cb) => query.run(peers, cb), maxTimeout)((err) => {
+        // combine peers from each path
+        paths.forEach((path) => {
+          path.toArray().forEach((peer) => {
+            out.push(peer)
```
vasco-santos: Same as reported above

src/query.js
```diff
@@ -40,22 +55,44 @@ class Query {
...
-      (cb) => PeerQueue.fromKey(this.key, cb),
-      (q, cb) => {
-        run.peersToQuery = q
-        each(peers, (p, cb) => addPeerToQuery(p, this.dht, run, cb), cb)
-      },
-      (cb) => workerQueue(this, run, cb)
-    ], (err) => {
+    // create correct number of paths
+    const numPaths = Math.min(c.DISJOINT_PATHS, peers.length)
+    const pathPeers = []
```
vasco-santos:
```js
const pathPeers = Array(numPaths).fill([])
```

vasco-santos: And remove the cycle below

src/index.js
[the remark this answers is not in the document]
jhiesey: Sure, can do.

src/index.js
jhiesey: Sure

src/private.js
jhiesey: Neither `out` nor `path` are arrays, so I don't see how the above solution applies here. What do you have in mind?

src/query.js
jhiesey: That won't work, since all of the array elements will be references to the same empty array.

src/index.js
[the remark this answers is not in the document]
jhiesey: Actually, I'm not sure where you're proposing to put it. It can't be after the return statement, or it won't get executed. Do you want the returned function assigned to a variable, then the push, and finally a separate return statement?

src/private.js
jhiesey: I could just replace them with arrays, and either just allow extra elements in the return value, or slice off the extras.

src/index.js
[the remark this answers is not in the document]
vasco-santos: I think you can do it before:

```js
cb(null, res)
```

that is, inside the return function. What do you think?

src/private.js
vasco-santos: If I am understanding correctly, `out` is an array and you are transforming `path` to an array. So, I think we can use the same `concat` logic as I suggested above. Am I missing something?

src/index.js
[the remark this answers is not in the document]
jhiesey: That would run the push once per query instead of once per path, so we would have lots of duplicates. Unless I'm misunderstanding?

src/private.js
jhiesey: `out` isn't an array. Are you saying to make it one?

src/index.js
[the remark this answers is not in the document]
vasco-santos: oh yes, you are right! It is ok then

src/private.js
vasco-santos: Ok, since it is a LimitedPeerList, we can leave it this way

jhiesey:
Go version: https://github.com/libp2p/go-libp2p-kad-dht/pull/204

src/constants.js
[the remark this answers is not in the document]
jhiesey: Yup.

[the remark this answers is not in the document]
jhiesey: Yep, done.

src/query.js
```diff
@@ -10,21 +10,38 @@ const PeerQueue = require('./peer-queue')
...
+
+  /**
+   * Query function.
+   * @typedef {queryFunc} function
+   * @param {PeerId} next - Peer to query
+   * @param {function(Error, Object)} callback - Query result callback
+   */
+
+  /**
+   * Create a new query. The makePath callback is called once per disjoint path, so that per-path
```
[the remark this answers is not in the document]
jhiesey: Sure. Done!

PR #40, opened 2018-09-20
fix: dht get options

vasco-santos:
Changed `dht.get` and `dht.getMany` `maxTimeout` parameter to an `options` parameter with the `maxTimeout` property as discussed [here](https://github.com/libp2p/js-libp2p/pull/248#pullrequestreview-156962993).


vasco-santos:
Thanks for the review @jacobheun ! I will restart the CI to get it green

PR #41, opened 2018-09-22
When a kad entry expires, use dsKey to remove from datastore

### LINKED ISSUE #38: Shouldn't the kad delete using the Key, not the buffer?
I'm getting errors on this line when I plug in a fs-datastore and data is > 1 day old.  I believe the "key" should be the dsKey

https://github.com/libp2p/js-libp2p-kad-dht/blob/edab8a3765d440c5c010c119a3c0a83ae37d6590/src/private.js#L136

blakebyrnes:
fix: use dskey instead of key for removal from datastore

Fixes #38


vasco-santos:
Thanks @blakebyrnes !

Can you just amend your commits to pass the commitlint CI job? 

Something like:
`fix: delete should use dskey for old records`
`chore: add tests`

PR #42, opened 2018-09-24
feat: start random walk and allow configuration for disabling

vasco-santos:
In the context of the integration of the DHT to `js-ipfs` [js-ipfs#856](https://github.com/ipfs/js-ipfs/pull/856).

I changed the implementation of `randomWalk`. It was using a `setInterval`, which was causing this function to run the `_walk` function after five minutes. On the contrary, `go-libp2p-kad-dht` runs immediately after the node starts and then, each 5 minutes, as we can see in [go-libp2p-kad-dht/dht_bootstrap.go#L81-L85](https://github.com/libp2p/go-libp2p-kad-dht/blob/master/dht_bootstrap.go#L81-L85)

Worklist:

- [x] Start `random-walk` once the DHT is started
- [x] Accept configuration for disabling the default behavior

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -115,14 +115,7 @@ class KadDHT {
    */
   start (callback) {
     this._running = true
-    this.network.start((err) => {
-      if (err) {
-        return callback(err)
-      }
-      this.randomWalk.start()
-
-      callback()
-    })
+    this.network.start(callback)
   }
 
   /**
@@ -134,7 +127,7 @@ class KadDHT {
    */
   stop (callback) {
     this._running = false
-    this.randomWalk.stop(() => {
+    this.randomWalk.stop(() => { // guarantee that random walk is stopped if it was started
       this.providers.stop()
       this.network.stop(callback)
     })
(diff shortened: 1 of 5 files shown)


src/index.js
```diff
@@ -115,7 +115,14 @@ class KadDHT {
    */
   start (callback) {
     this._running = true
-    this.network.start(callback)
+    this.network.start((err) => {
+      if (err) {
+        return callback(err)
+      }
+      this.randomWalk.start()
```
diasdavid: This should be started just like any other Peer Discovery Service, that means, start it on js-libp2p instead.

vasco-santos: I will change the PR and create a new one for `js-libp2p`! Thanks

vasco-santos:
I will look into that problem @jacobheun . Thanks for finding it out

vasco-santos:
@jacobheun it is green now!

@diasdavid can we have your feedback to move this forward?
Explanation in this [discussion](https://github.com/libp2p/js-libp2p/pull/251#discussion_r220654190)

PR #51, opened 2018-11-08
feat: select first record when no selector function

vasco-santos:
Select first record when no selector function exists for it. Currently, it throws an handled exception and `go-ipfs` allows other types of keys.

### DIFF
--- a/src/private.js
+++ b/src/private.js
@@ -271,7 +271,17 @@ module.exports = (dht) => ({
       (cb) => dht.getMany(key, 16, options.maxTimeout, cb),
       (vals, cb) => {
         const recs = vals.map((v) => v.val)
-        const i = libp2pRecord.selection.bestRecord(dht.selectors, key, recs)
+        let i = 0
+
+        try {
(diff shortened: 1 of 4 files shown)


src/private.js
```diff
@@ -271,7 +271,17 @@ module.exports = (dht) => ({
...
         const recs = vals.map((v) => v.val)
-        const i = libp2pRecord.selection.bestRecord(dht.selectors, key, recs)
+        let i = 0
+
+        try {
+          i = libp2pRecord.selection.bestRecord(dht.selectors, key, recs)
+        } catch (err) {
+          // Assume the first record if no selector available
+          if (err.code !== 'ERR_NO_SELECTOR_FUNCTION_FOR_RECORD_KEY') {
+            return callback(err)
```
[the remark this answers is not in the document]
vasco-santos: right, it is better to have that test now than have tests failing later. Will add it now.

jacobheun: Call `cb` and let waterfall handle calling `callback`
```suggestion
            return cb(err)
```

vasco-santos: thanks @jacobheun

PR #53, opened 2018-11-12
chore: add error code to all errors

vasco-santos:
Added `err-code` to all the errors, in order to improve debuggability for `libp2p-kad-dht`


src/network.js
```diff
@@ -38,12 +40,12 @@ class Network {
...
 
     if (this.isOnline) {
-      return cb(new Error('Network is already running'))
+      return cb(errcode(new Error('Network is already running'), 'ERR_NETWORK_ALREADY_RUNNING'))
     }
 
     // TODO add a way to check if switch has started or not
     if (!this.dht.isStarted) {
-      return cb(new Error('Can not start network'))
+      return cb(errcode(new Error('Can not start network'), 'ERR_CANNOT_START_NETWORK'))
```
jacobheun: Keep this in line with the other network errors.
```suggestion
      return cb(errcode(new Error('Can not start network'), 'ERR_NETWORK_CANNOT_START'))
```

vasco-santos:
@jacobheun I think that I made a good first pass with tests

src/index.js
```diff
@@ -266,24 +267,25 @@ class KadDHT {
...
+                if (!(err.code === 'ERR_INVALID_RECORD')) {
                   return cb(err)
                 }
               }
 
               const res = { closerPeers: peers }
 
-              if ((rec && rec.value) ||
-                  err instanceof errors.InvalidRecordError) {
+              if ((rec && rec.value) || err.code === 'ERR_INVALID_RECORD') {
```
jacobheun: Need to check the existence of err. If there is no record/record value and no error, this would break.
```suggestion
              if ((rec && rec.value) || (err && err.code === 'ERR_INVALID_RECORD')) {
```

src/network.js
```diff
@@ -222,14 +224,14 @@ function writeReadMessage (conn, msg, callback) {
...
-        return callback(new Error('No message received'))
+        return callback(errcode(new Error('No message received'), 'ERR_NO_MESSAGE_RECEIVED'))
       }
 
       let response
       try {
         response = Message.deserialize(res[0])
       } catch (err) {
-        return callback(new Error('failed to deserialize response: ' + err.message))
+        return callback(errcode(new Error(`failed to deserialize response: ${err.message}`), 'ERR_FAILED_DESERIALIZE_RESPONSE'))
```
jacobheun: There's a few places where we're losing the call stack by recreating the error. Since we're using error codes now, I think losing the nicer message is worth keeping the call stack. 
```suggestion
        return callback(errcode(err, 'ERR_FAILED_DESERIALIZE_RESPONSE'))
```

PR #57, opened 2018-11-29
feat: allow configurable validators and selectors

vasco-santos:
Allow configurable validators and selectors. Users of `libp2p-kad-dht` must be able to set their own validators and selectors. 

For instance, this is important for `js-ipfs` to use `/ipns/` namespaces in the DHT

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -36,10 +36,14 @@ class KadDHT {
    * @param {number} options.kBucketSize k-bucket size (default 20)
    * @param {Datastore} options.datastore datastore (default MemoryDatastore)
    * @param {boolean} options.enabledDiscovery enable dht discovery (default true)
+   * @param {object} options.validators validators object with namespace as keys and function(key, record, callback)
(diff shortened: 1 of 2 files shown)


src/index.js
```diff
@@ -36,10 +36,14 @@ class KadDHT {
...
    * @param {Datastore} options.datastore datastore (default MemoryDatastore)
    * @param {boolean} options.enabledDiscovery enable dht discovery (default true)
+   * @param {object} options.validators validators object with namespace as keys and function(key, record, callback)
+   * @param {object} options.selectors selectors object with namespace as keys and function(key, records)
    */
   constructor (sw, options) {
     assert(sw, 'libp2p-kad-dht requires a instance of Switch')
     options = options || {}
+    options.validators = options.validators || {}
+    options.selectors = options.selectors || {}
```
jacobheun: This is unnecessary isn't it? The spread operator will ignore `undefined` and `null`.

vasco-santos: Just tested it. You are right, removed the fallback 😀

PR #62, opened 2018-12-11
chore: update options timeout property

vasco-santos:
`maxTimeout ` option property renamed to `timeout`. Moreover, previous support for timeout using a number instead of an object with options deprecated.

BREAKING CHANGE: get, getMany, findProviders and findPeer do not accept a timeout number anymore. It must be a property of an object options.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -203,7 +203,7 @@ class KadDHT {
    *
    * @param {Buffer} key
    * @param {Object} options - get options
-   * @param {number} options.maxTimeout - optional timeout (default: 60000)
+   * @param {number} options.timeout - optional timeout (default: 60000)
    * @param {function(Error, Buffer)} callback
    * @returns {void}
    */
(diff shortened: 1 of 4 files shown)


vasco-santos:
coverage decreases as we are still supporting `maxTimeout` until the next release.

src/private.js
```diff
@@ -458,14 +458,14 @@ module.exports = (dht) => ({
    * Search the dht for up to `n` providers of the given CID.
    *
    * @param {CID} key
-   * @param {number} maxTimeout - How long the query should maximally run in milliseconds.
+   * @param {number} provideTimeout - How long the query should maximally run in milliseconds.
```
fsdiogo:
```suggestion
   * @param {number} providerTimeout - How long the query should maximally run in milliseconds.
```

src/private.js
```diff
@@ -458,14 +458,14 @@ module.exports = (dht) => ({
...
    * @param {number} n
    * @param {function(Error, Array<PeerInfo>)} callback
    * @returns {void}
    *
    * @private
    */
-  _findNProviders (key, maxTimeout, n, callback) {
+  _findNProviders (key, provideTimeout, n, callback) {
```
fsdiogo:
```suggestion
  _findNProviders (key, providerTimeout, n, callback) {
```

src/private.js
```diff
@@ -524,7 +524,7 @@ module.exports = (dht) => ({
 
       const peers = dht.routingTable.closestPeers(key.buffer, c.ALPHA)
 
-      timeout((cb) => query.run(peers, cb), maxTimeout)((err) => {
+      timeout((cb) => query.run(peers, cb), provideTimeout)((err) => {
```
fsdiogo:
```suggestion
      timeout((cb) => query.run(peers, cb), providerTimeout)((err) => {
```

src/index.js
```diff
@@ -232,24 +230,22 @@ class KadDHT {
...
-        maxTimeout: options
-      }
     } else {
       options = options || {}
     }
 
-    if (!options.maxTimeout) {
-      options.maxTimeout = c.minute
+    if (!options.maxTimeout && !options.timeout) {
+      options.timeout = 'c.minute' // default
```
alanshaw: Is this meant to be a string?

PR #69, opened 2019-01-03
chore: add basic usage example

### LINKED ISSUE #36: Usage Example
I couldn't find a clear usage example/documentation. Can anyone help me out?

chris-asl:
@vasco-santos This is an initial attempt on the usage example for issue #36.
PS: I don't know whether you're signing your commits in libp2p, I can remove it.
Same goes for any other changes :-) 

----------------------------------
Add a basic usage example whereby the client code runs on top of each
libp2p DHT node. The idea is that each user has its own deaddrop,
identified by her UserId (on the DHT identifiers namespace), where other
users can leave messages to.

The demo instantiates three application (and thus DHT) nodes. Bootstrap
is handled by having an "always-on" node where others can connect to.
In real world examples, this would not suffice, as a discovery mechanism
needs to be in place (either by using the RandomWalk peer discovery or
by switching any implementations of libp2p/interface-peer-discovery).

Once the nodes are instantiated and connected, thus forming a single
DHT, one deaddrop application (userA) writes a message for userB. Then,
userB reads her messages (after having been informed about the message
through off-channel means).

For this example to make sense, any sent message is encrypted using the
recipient's public key, so that only she can read them, by decrypting
them with her private key.

Resolves #36.

Signed-off-by: Chris Aslanoglou <chris.aslanoglou@gmail.com>

### DIFF
--- a/examples/deaddrops/deaddrop.js
+++ b/examples/deaddrops/deaddrop.js
@@ -1,0 +1,58 @@
+'use strict'
+
+const waterfall = require('async').waterfall
+
+const DHTNode = require('./dht').DHTNode
+
+class DeadDrop {
+	constructor(userId, dht) {
+		this.userId = userId
+		this.userIdKey = Buffer.from(userId)
+		this.dht = dht
+	}
+
+	init(cb) {
+		this.dht.start(cb)
+	}
+
+	dial(target, cb) {
+		this.dht.dial(target, cb)
+	}
+
+	shutdown(cb) {
+		this.dht.stop(cb)
+	}
+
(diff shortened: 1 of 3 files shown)


vasco-santos:
Thanks for the PR @chris-asl and sorry for the late response.

We are currently on a process to refactor this codebase to use `async await` and `async iterators / generators`, and consequently I will wait for that refactor before merging the example, as it will need to be updated.

achingbrain:
Thanks for opening this, I'm sorry it never landed. The codebase has been through several refactors and this may be better off being started again if the examples are still useful.

PR #71, opened 2019-01-04
chore: update dependencies

vasco-santos:


jacobheun:
Code coverage change is nominal and the only line changes are for tests. This should be good to go.

vasco-santos:
I agree! Will merge and release it! Thanks @jacobheun

Issue #68, opened 2018-12-28
During a put call, if any node fails to put, the entire put fails


blakebyrnes:
Have a scenario where if you get closest peers, and one peer cannot reply back with success, the entire put fails.  I think ideally, if X number of peers can be written to, this should succeed.

Examples of what a node might not be able to write to kad:
1) The KAD is not initialized on that node.  Ie, it's running a "light" node.  Solutions to this could include an earlier detection (maybe in identify) of the protocols that a given node supports - and then not adding them to the kad peer group if the node does not support /ipfs/kad/1.0.0
2) The node is no longer connected

I believe the simplest solution is to change the "each" in async handler to a "some" handler, and changing the variable name of cb to something different for the inner loop (just a best practice).

  ```
put (key, value, callback) {
    this._log('PutValue %b', key)

    waterfall([
      (cb) => utils.createPutRecord(key, value, cb),
      (rec, cb) => waterfall([
        (cb) => this._putLocal(key, rec, cb),
        (cb) => this.getClosestPeers(key, cb),
        (peers, cb) => **each**(peers, (peer, **cb**) => {
          this._putValueToPeer(key, rec, peer, **cb**)
        }, cb)
      ], cb)
    ], callback)
  }
```
@jacobheun @vasco-santos  I can create a pr if you agree with the approach.

vasco-santos:
Hello @blakebyrnes and thanks for reaching out and sorry for the delay answering this! 

I agree that we should not fail if we are able to put to a good set of peers and failing one for instance. This way, I believe that the best option here is to have a parameter to set the minimum number of peers to have a successful put (not applicable if the number of closest peers is less than it).

What do you think? It would be great if you could create a PR with this.

blakebyrnes:
Sounds good.  I've got an implementation of this going.  Are you ok with the param being called "minPutPeers"?  I'm looking into tests right now, and several tests around "put" are failing with the default of 1 put peer that I chose.  Do you think we should have the default be 0 if no minimum peers are provided?  It would keep the behavior the same as current, but I do think it's potentially unexpected that the system is ok with putting to no peers.

jacobheun:
I'd recommend a generic option like `minPeers`, since it might be good to reuse that elsewhere in the future and the called function already declares the `put`.

I also recommend keeping the current behavior, default of 0. While this may be unexpected behavior it prevents this from being a breaking change. In the future we can then change the behavior to what we would expect.

[...]

PR #72, opened 2019-01-08
fix: specify # of peers for successful put

blakebyrnes:
put call fails if any node fails for any reason - including disconnected, node doesn’t support kad protocol, etc.  Changed behavior to be able to provide a minimum number of peers to write to to be considered a “successful put”.  Default put peers required was chosen as 0 so as to be a non-breaking change

Fix for #68

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -4,6 +4,7 @@ const libp2pRecord = require('libp2p-record')
 const MemoryStore = require('interface-datastore').MemoryDatastore
 const waterfall = require('async/waterfall')
 const each = require('async/each')
+const filter = require('async/filter')
 const timeout = require('async/timeout')
 const PeerId = require('peer-id')
 const PeerInfo = require('peer-info')
@@ -181,7 +182,7 @@ class KadDHT {
    * @param {Buffer} value
    * @param {Object} options - get options
    * @param {number} options.minPeers - minimum peers that must be put to to consider this a successful operation
-   * (default: 0)
+   * (default: closestPeers.length)
    * @param {function(Error)} callback
    * @returns {void}
    */
@@ -192,42 +193,41 @@ class KadDHT {
     } else {
       options = options || {}
     }
-    if (!options.minPeers) {
-      options.minPeers = 0 // default
-    }
-    this._log('PutValue %b', key)
 
-    let counter = 0
+    this._log('PutValue %b', key)
 
     waterfall([
       (cb) => utils.createPutRecord(key, value, cb),
       (rec, cb) => waterfall([
         (cb) => this._putLocal(key, rec, cb),
         (cb) => this.getClosestPeers(key, cb),
-        (peers, cb) => each(peers, (peer, cb) => {
-          this._putValueToPeer(key, rec, peer, (err) => {
-            if (err) {
-              this._log.error('Failed to put to peer (%b): %s', peer.id, err)
-            } else {
-              counter += 1
+        (peers, cb) => {
+          // Ensure we have a default `minPeers`
+          options.minPeers = options.minPeers || peers.length
+          // filter out the successful puts
+          filter(peers, (peer, cb) => {
+            this._putValueToPeer(key, rec, peer, (err) => {
+              if (err) {
+                this._log.error('Failed to put to peer (%b): %s', peer.id, err)
+                return cb(null, false)
+              }
+              cb(null, true)
+            })
+          }, (err, results) => {
+            if (err) return cb(err)
+
+            // Did we put to enough peers?
+            if (options.minPeers > results.length) {
+              const error = errcode(new Error('Failed to put value to enough peers'), 'ERR_NOT_ENOUGH_PUT_PEERS')
+              this._log.error(error)
+              return cb(error)
             }
-            // always ok result
+
             cb()
           })
-        }, cb)
+        }
       ], cb)
-    ], (err) => {
-      if (err) {
-        return callback(err)
-      }
-      if (counter < options.minPeers) {
-        const errMsg = 'Failed to put value to enough peers'
-
-        this._log.error(errMsg)
-        return callback(errcode(new Error(errMsg), 'ERR_NOT_ENOUGH_PUT_PEERS'))
-      }
-      callback()
-    })
+    ], callback)
   }
 
   /**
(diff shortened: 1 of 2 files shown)


src/index.js
```diff
@@ -179,22 +179,55 @@ class KadDHT {
    *
    * @param {Buffer} key
    * @param {Buffer} value
+   * @param {Object} options - get options
+   * @param {number} options.minPeers - minimum peers that must be put to to consider this a successful operation
+   * (default: 0)
```
vasco-santos: We cannot have this default with this logic. When this was suggested https://github.com/libp2p/js-libp2p-kad-dht/issues/68#issuecomment-452390478, it was to keep the current behaviour as default. That is, the default should be the number of closestPeers.

I believe that what jacob was suggesting was that we should keep the default of 0 fails to put.

blakebyrnes: Right now, the behavior succeeds if you find NO peers to put to.  It will fail if you request a put to any peer, and it fails.

The way the logic is written, if you put 0 (or put nothing), it will not fail if you fail to put to any peers.  It will continue to fail as it did before if any individual peer put fails.  

If you put 1, the logic now requires at least 1 successful put.  No peers will fail, as will finding 3 peers, each of which fails.

Is this what you were thinking it did?  Or is the variable just poorly named?

blakebyrnes: @jacobheun did you by chance look at this pr?  do you remember your comment about minPeers?   happy to modify with your or @vasco-santos advice.  just let me know!

jacobheun: Ok, I previously misunderstood the logic, we'll need to update this PR.

**Current functionality**
1. Get the closest peers
1. For each peer, put to that peer.
1. If **any** put fails, the request fails.

So the current logic is that `minPeers === closestPeers.length`. So once we get the closest peers, if `minPeers` is not set, it should be set to the number of peers returned.

blakebyrnes:
Ok, I follow now.  Sorry.  The naming of these variables was throwing me into a mental loop.  Should be fixed now - added a test to confirm

jacobheun:
This could probably be a bit simpler with `async/filter`. Here's a working diff:

```diff
diff --git a/src/index.js b/src/index.js
index eb4848d..59d56a8 100644
--- a/src/index.js
+++ b/src/index.js
@@ -4,6 +4,7 @@ const libp2pRecord = require('libp2p-record')
 const MemoryStore = require('interface-datastore').MemoryDatastore
 const waterfall = require('async/waterfall')
 const each = require('async/each')
+const filter = require('async/filter')
 const timeout = require('async/timeout')
 const PeerId = require('peer-id')
 const PeerInfo = require('peer-info')
@@ -181,7 +182,7 @@ class KadDHT {
    * @param {Buffer} value
    * @param {Object} options - get options
    * @param {number} options.minPeers - minimum peers that must be put to to consider this a successful operation
-   * (default: 0)
+   * (default: closestPeers.length)
    * @param {function(Error)} callback
    * @returns {void}
    */
@@ -195,47 +196,39 @@ class KadDHT {
 
     this._log('PutValue %b', key)
 
-    let failedPuts = 0
-    let successfulPuts = 0
-
     waterfall([
       (cb) => utils.createPutRecord(key, value, cb),
       (rec, cb) => waterfall([
         (cb) => this._putLocal(key, rec, cb),
         (cb) => this.getClosestPeers(key, cb),
-        (peers, cb) => each(peers, (peer, cb) => {
-          this._putValueToPeer(key, rec, peer, (err) => {
-            if (err) {
-              this._log.error('Failed to put to peer (%b): %s', peer.id, err)
-              failedPuts += 1
-            } else {
-              successfulPuts += 1
+        (peers, cb) => {
+          // Ensure we have a default `minPeers`
+          options.minPeers = options.minPeers || peers.length
+
+          // filter out the successful puts
+          filter(peers, (peer, _cb) => {
+            this._putValueToPeer(key, rec, peer, (err) => {
+              if (err) {
+                this._log.error('Failed to put to peer (%b): %s', peer.id, err)
+                return _cb(null, false)
+              }
+              _cb(null, true)
+            })
+          }, (err, results) => {
+            if (err) return cb(err)
+
+            // Did we put to enough peers?
+            if (options.minPeers > results.length) {
+              const error = errcode(new Error('Failed to put value to enough peers'), 'ERR_NOT_ENOUGH_PUT_PEERS')
+              this._log.error(error)
+              return cb(error)
             }
-            // always ok result
+
             cb()
           })
-        }, cb)
-      ], cb)
-    ], (err) => {
-      if (err) {
-        return callback(err)
-      }
-
-      if (isNaN(options.minPeers)) {
-        if (failedPuts > 0) {
-          const errMsg = 'Failed to put value to a peer'
-
-          this._log.error(errMsg)
-          return callback(errcode(new Error(errMsg), 'ERR_PUT_FAILED'))
         }
-      } else if (successfulPuts < options.minPeers) {
-        const errMsg = 'Failed to put value to enough peers'
-
-        this._log.error(errMsg)
-        return callback(errcode(new Error(errMsg), 'ERR_NOT_ENOUGH_PUT_PEERS'))
-      }
-      callback()
-    })
+      ], cb)
+    ], callback)
   }
 
   /**
diff --git a/test/kad-dht.spec.js b/test/kad-dht.spec.js
index d9c2a98..6b06a72 100644
--- a/test/kad-dht.spec.js
+++ b/test/kad-dht.spec.js
@@ -374,7 +374,7 @@ describe('KadDHT', () => {
         }
       ], (err) => {
         expect(err).to.exist()
-        expect(err.code).to.eql('ERR_PUT_FAILED')
+        expect(err.code).to.eql('ERR_NOT_ENOUGH_PUT_PEERS')
         stub.restore()
         tdht.teardown(done)
       })

```

blakebyrnes:
Way cleaner. I re-pushed with your changes.

src/index.js
```diff
@@ -191,7 +202,29 @@ class KadDHT {
         (cb) => this._putLocal(key, rec, cb),
         (cb) => this.getClosestPeers(key, cb),
         (peers, cb) => each(peers, (peer, cb) => {
```
jacobheun: I forgot to get rid of the each in my diff, it needs to get removed. It can just be:
```suggestion
        (peers, cb) => {
```

blakebyrnes: Somehow missed that.  Meant to remove it.

Issue #76, opened 2019-02-01
Allow to set randomWalk parameter through config


thomaseizinger:
Currently, the `randomWalk` discovery is started when the DHT module is started:

https://github.com/libp2p/js-libp2p-kad-dht/blob/b2a9fc5ab53db5a0105004e30af29fb24beaddfe/src/index.js#L136-L147

However, because no parameters are passed to `randomWalk.start()`, the random walk starts with the default parameters:

https://github.com/libp2p/js-libp2p-kad-dht/blob/b2a9fc5ab53db5a0105004e30af29fb24beaddfe/src/random-walk.js#L26-L34

I think those parameters should be exposed in the config as something like:

```js

dht: {
    kBucketSize: 20,
    randomWalk: {
       enabled: true // This could replace the current `enabledDiscovery` flag,
       queriesPerPeriod: 1,
       interval: 30000,
       timeout: 10000
    }
}
```

a1300:
I second this. This is missing for me also.

vasco-santos:
Totally agree with this! Thanks for opening the issue and propose a solution @thomaseizinger

jhiesey:
Does the random walk logic really belong in the DHT at all? Ultimately running the random walk just calls `findPeer`, which is public.

thomaseizinger:
Are you suggesting that the functionality of random walks should be extracted into its own library - using what provides the ability to `findPeer`?

I think it could make sense.
However, for the time being, it is included in this one but wasn't configurable from the outside, so I'd suggest that this is actually a different topic :)

vasco-santos:
We may consider having the `randomWalk` in a separate module when we refactor the DHT. For now, I think this is ok, but we can create an issue for that. Meanwhile, I will close this one.

Thanks for your contributions!

PR #77, opened 2019-02-01
Expose randomWalk parameters in config

thomaseizinger:
Implementation of what is suggested in #76 since I need it for a PoC. Happy to revise if it should be done differently :)

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -46,6 +46,7 @@ class KadDHT extends EventEmitter {
     options = options || {}
     options.validators = options.validators || {}
     options.selectors = options.selectors || {}
+    options.randomWalk = options.randomWalk || {}
 
     /**
      * Local reference to the libp2p-switch instance
@@ -118,6 +119,9 @@ class KadDHT extends EventEmitter {
      * Random walk state, default true
      */
     this.randomWalkEnabled = !options.hasOwnProperty('enabledDiscovery') ? true : Boolean(options.enabledDiscovery)
+    this.randomWalkQueriesPerPeriod = !options.randomWalk.hasOwnProperty('queriesPerPeriod') ? RandomWalk.prototype.defaultQueriesPerPeriod : parseInt(options.randomWalk.defaultQueriesPerPeriod)
+    this.randomWalkInterval = !options.randomWalk.hasOwnProperty('interval') ? RandomWalk.prototype.defaultInterval : parseInt(options.randomWalk.interval)
+    this.randomWalkTimeout = !options.randomWalk.hasOwnProperty('timeout') ? RandomWalk.prototype.defaultTimeout : parseInt(options.randomWalk.timeout)
   }
 
   /**
@@ -143,7 +147,7 @@ class KadDHT extends EventEmitter {
       }
 
       // Start random walk if enabled
-      this.randomWalkEnabled && this.randomWalk.start()
+      this.randomWalkEnabled && this.randomWalk.start(this.randomWalkQueriesPerPeriod, this.randomWalkInterval, this.randomWalkTimeout)
       callback()
     })
   }
--- a/src/index.js
+++ b/src/index.js
@@ -22,6 +22,7 @@ const Providers = require('./providers')
 const Message = require('./message')
 const RandomWalk = require('./random-walk')
 const assert = require('assert')
+const defaultsDeep = require('@nodeutils/defaults-deep')
 
 /**
  * A DHT implementation modeled after Kademlia with S/Kademlia modifications.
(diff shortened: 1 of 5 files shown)


thomaseizinger:
I just realized that such a change also needs to be incorporated here:

https://github.com/libp2p/js-libp2p/blob/c4cab007af5e7ec170615941230c18c893eece8a/src/config.js#L34-L40

I'll pause here then and wait for feedback to whether this feature is actually desired.

src/index.js
```diff
@@ -116,6 +117,9 @@ class KadDHT {
      * Random walk state, default true
      */
     this.randomWalkEnabled = !options.hasOwnProperty('enabledDiscovery') ? true : Boolean(options.enabledDiscovery)
```
vasco-santos: let's change the line above for what you proposed in the issue

src/random-walk.js
```diff
@@ -164,4 +164,8 @@ class RandomWalk {
   }
 }
 
+RandomWalk.prototype.defaultQueriesPerPeriod = 1
```
vasco-santos: Please, add these to to `constants` file instead

src/index.js
```diff
@@ -44,6 +44,7 @@ class KadDHT {
     options = options || {}
     options.validators = options.validators || {}
     options.selectors = options.selectors || {}
+    options.randomWalk = options.randomWalk || {}
```
vasco-santos: it would be cleaner to have the defaults here merged with what is received.

src/random-walk.js
```diff
@@ -29,9 +29,9 @@ class RandomWalk {
    * @returns {void}
    */
   start (queries, period, timeout) {
```
vasco-santos: Please remove the lines below and change the above for using the default parameters, as follows:

```js
start (queries = ..., )
```

I know the function was not implemented by you, but feels like a good oportunity to change this.

thomaseizinger:
https://github.com/libp2p/js-libp2p/pull/317 just got merged :tada: 

How are defaults handled across libp2p modules?
I have the feeling it would be better if a module like `js-libp2p-kad-dht` is unopinionated about defaults and *requires* all the config options to be set.

src/index.js
```diff
@@ -115,7 +116,10 @@ class KadDHT {
     /**
      * Random walk state, default true
      */
-    this.randomWalkEnabled = !options.hasOwnProperty('enabledDiscovery') ? true : Boolean(options.enabledDiscovery)
+    this.randomWalkEnabled = !options.randomWalk.hasOwnProperty('enabled') ? c.defaultRandomWalk.enabled : Boolean(options.randomWalk.enabled)
+    this.randomWalkQueriesPerPeriod = !options.randomWalk.hasOwnProperty('queriesPerPeriod') ? c.defaultRandomWalk.queriesPerPeriod : parseInt(options.randomWalk.queriesPerPeriod)
+    this.randomWalkInterval = !options.randomWalk.hasOwnProperty('interval') ? c.defaultRandomWalk.interval : parseInt(options.randomWalk.interval)
+    this.randomWalkTimeout = !options.randomWalk.hasOwnProperty('timeout') ? c.defaultRandomWalk.timeout : parseInt(options.randomWalk.timeout)
```
vasco-santos: I would recommend using [nodeutils/defaults-deep](https://github.com/nodeutils/defaults-deep#readme) in here, and in

```js
options.randomWalk = options.randomWalk || c.defaultRandomWalk
```

With the current code, if we receive a parameter as follows:

```
{
  enabled: true
}
```

The initial options object will only have that, as it exists. With my proposal, you can remove all the lines above here and default in the options above.

vasco-santos:
@thomaseizinger I want to have the defaults here too, as this implementation may be used by other projects, and it is better to fallback to a value than to require all the parameters.

thomaseizinger:
@vasco-santos sorry for the delay!

The code now uses `defaultsDeep`! Please have another look if you can spot anything else :)

vasco-santos:
@thomaseizinger thanks for the updates!

Basically, we are moving our CI from jenkins to travis now. So, the master branch already is using travis, but this PR came from an older version of it. I just rebased your PR with that to use travis as well! I will have a look to the PR!

src/index.js
```diff
@@ -46,6 +47,7 @@ class KadDHT extends EventEmitter {
     options = options || {}
     options.validators = options.validators || {}
     options.selectors = options.selectors || {}
+    options.randomWalk = defaultsDeep(options.randomWalk, c.defaultRandomWalk)
```
vasco-santos: Add `options.randomWalk` to the js docs above and describe the options type for it as follows:

```js
/**
 * Random walk options
 * @typedef {Object} randomWalk
 * @property {boolean} enabled - discovery enabled (default: true)
 * @property {number} queriesPerPeriod - how many queries to run per period (default: 1)
 * @property {number} interval - how often to run the the random-walk process, in milliseconds (default: 300000)
 * @property {number} timeout - how long to wait for the the random-walk query to run, in milliseconds (default: 10000)
*/
```

Also remove `options.enabledDiscovery` from the docs

thomaseizinger:
@vasco-santos I wasn't sure how you exactly meant the docs to be fixed so I took a guess at it :D

vasco-santos:
It was a good guess then 🎉 

Thanks for the PR @thomaseizinger

PR #82, opened 2019-02-21
WIP callbacks -> async / await

dirkmc:
BREAKING CHANGE: This change replaces callbacks with async / await

I refactored a few things while I was making this change:
- queue.js
  - functionally equivalent, but now uses Iterators instead of an async/queue
  - stops any running workers if an error is thrown
  - has a mechanism to stop a query once the required number of results have been fetched
  - ensure peers are unique by peer id (before they were only unique by Object reference)
- providers.js
  - now ensures that `addProvider()`, `getProviders()` and `_cleanup()` are synchronized (using a p-queue)
  - now cleans up any expired entries (previously expired entries would only be cleaned up if all peer ids for a given CID had expired)
  - add items to the cache any time we go to the datastore (previously we would cache on `addProvider()` but not on `getProviders()`)
- random-walk.js
  - simplify scheduling mechanism
  - fix a bug where we [check a non-existent variable on start](https://github.com/libp2p/js-libp2p-kad-dht/blob/master/src/random-walk.js#L33)
- getMany()
  - Fix bug where requesting a single value would fail if the value was not found in the local datastore (ie before it wouldn't go out to the DHT to look for it. Now it does)
- stop()
  - Ensure all running queries are stopped when the DHT is stopped
- Moved all pull-streams into a single class so that we can easily replace with Iterators when the `Connection` interface has been updated


package.json
```diff
@@ -45,14 +45,15 @@
     "err-code": "^1.1.2",
     "hashlru": "^2.3.0",
     "heap": "~0.2.6",
-    "interface-datastore": "~0.6.0",
+    "interface-datastore": "ipfs/interface-datastore#refactor/async-iterators",
```
vasco-santos: remove `async` dependency?

src/index.js
```diff
@@ -253,145 +225,133 @@ class KadDHT extends EventEmitter {
...
-        return callback(null, vals)
-      }
+    if (!err) {
+      vals.push({
+        val: localRec.value,
+        from: this.peerInfo.id
+      })
+    }
+
+    if (nvals <= 1) {
```
vasco-santos: Do not forget to change this as we discussed :)

src/index.js
```diff
@@ -457,25 +412,22 @@ class KadDHT extends EventEmitter {
...
-
-        each(peers, (peer, cb) => {
-          this._log('putProvider %s to %s', key.toBaseEncodedString(), peer.toB58String())
-          this.network.sendMessage(peer, msg, cb)
-        }, cb)
-      }
-    ], (err) => callback(err))
+    await this.providers.addProvider(key, this.peerInfo.id)
+
+    const peers = await this.getClosestPeers(key.buffer)
```
vasco-santos: What about doing these 2 before in parallel?

src/index.js
```diff
@@ -535,80 +471,69 @@ class KadDHT extends EventEmitter {
 
     this._log('findPeer %s', id.toB58String())
 
-    this.findPeerLocal(id, (err, pi) => {
-      if (err) {
-        return callback(err)
-      }
+    const pi = await this.findPeerLocal(id)
```
vasco-santos: can we call this `pid` our `peerId` instead? I know that was before, but `pi` seems odd

dirkmc:
Yes I think that makes sense 👍

kumavis:
I don't mean to disparage the great effort here but in my experience large on-going refactors are a bad idea when working with multiple contributors. You end up in merge conflict hell and in resolving them its easy to run over patches that were merged after the refactor started.

Also seems this one has also outgrown its original async/await focus and has become a something more, possibly now a sort of too-big-to-fail PR that blocks or at least complicates all development. 

I don't mean to sound dire, I've just lived through too many of these. And I'm mostly an outsider here so I'm not trying to tell someone what to do but I'd recommend (1) resolving and merging this asap (2) using the code here to break this down into multiple smaller PRs 

happy to do my part and not just comment : )

I anticipate working very closely with the dht this month

src/query.js
```diff
@@ -1,242 +1,393 @@
...
+   */
+  async * run (peers, timeout) {
     if (peers.length === 0) {
       this._log.error('Running query with no peers')
-      return callback()
+      return
+    }
+
+    const run = new Run(this.dht, this.key, this.makePathQuery, this._log)
+    this.run = run
```
kumavis: seems like `this.run` accidentally clashes with the `run` function defined on the `Query` class. I guess this could be an intentional override, but if so it should be heavily commented because its surprising

src/query.js
```diff
@@ -1,242 +1,393 @@
...
+  }
+
+  /**
+   * Stop the Query if it is running. This should be called after run() if the
+   * Query hasn't completed (eg if the caller only needs a limited number of
+   * results) so that the Query doesn't continue traversing the DHT
+   * unnecessarily.
+   */
+  stop () {
+    this.run && this.run.stop()
```
kumavis: this seems like a mistake (as I elsewhere noted about `this.run`). I would expect `this.run` to always be truthy because `run` is defined on the class

package.json
```diff
@@ -38,26 +38,27 @@
...
-    "async": "^2.6.2",
     "base32.js": "~0.1.0",
     "chai-checkmark": "^1.0.1",
     "cids": "~0.5.7",
     "debug": "^4.1.1",
     "err-code": "^1.1.2",
     "hashlru": "^2.3.0",
     "heap": "~0.2.6",
-    "interface-datastore": "~0.6.0",
+    "interface-datastore": "ipfs/interface-datastore#refactor/async-iterators",
```
kumavis: issue with `interface-datastore` dep ref
```
➜ npm i
npm ERR! code 1
npm ERR! Command failed: git checkout refactor/async-iterators
npm ERR! error: pathspec 'refactor/async-iterators' did not match any file(s) known to git
```

kumavis: looks like merged and branch deleted :+1:

kumavis: not published yet though, I'll point to the commit for now

package.json
```diff
@@ -38,26 +38,27 @@
...
     "k-bucket": "^5.0.0",
-    "libp2p-crypto": "~0.16.0",
-    "libp2p-record": "~0.6.2",
+    "libp2p-crypto": "libp2p/js-libp2p-crypto#feat/async-await",
+    "libp2p-record": "dirkmc/js-libp2p-record#feat/async-await",
     "multihashes": "~0.4.14",
-    "multihashing-async": "~0.5.2",
-    "peer-id": "~0.12.2",
-    "peer-info": "~0.15.1",
+    "multihashing-async": "multiformats/js-multihashing-async#feat/async-iterators",
```
kumavis: published as `0.7.0` :+1:

kumavis:
Theres some git branch dep hell going on as async/await branches gets merged deleted

kumavis:
related PRs in deps
- [ ] https://github.com/libp2p/js-libp2p-crypto/pull/131
- [ ] https://github.com/dirkmc/js-libp2p-record/tree/feat/async-await (no PR?)
- [ ] https://github.com/libp2p/js-peer-id/pull/87
- [ ] https://github.com/libp2p/js-peer-info/pull/67

vasco-santos:
Thanks for your feedback @kumavis 

We will need to update this PR with the new changes, and we will do it while the remaining dependencies are not ready. Hopefully, we will be able to get this PR more ready to merge in the next month

src/query.js
kumavis: @dirkmc could you clarify if this is intentional or not

dirkmc: @kumavis it's not intentional - thanks for pointing it out.

I think it's best to hold off on updating this PR until the dependencies are released

dirkmc:
@kumavis if you are interested in helping out with DHT work it would be very welcome. This issue is one we will need to implement in both Go and Javascript: https://github.com/libp2p/go-libp2p-kad-dht/issues/323

kumavis:
@dirkmc hey/ took a stab at adding your refactors in a non-breaking way to reduce conflicts https://github.com/libp2p/js-libp2p-kad-dht/pull/108

src/query.js
kumavis: opened an issue https://github.com/libp2p/js-libp2p-kad-dht/issues/114

kumavis:
@dirkmc theres a lot of great work in here that i dont want to get lost, but extracting the high level changes from the diff is a bit difficult -- if there are any additional fixes or changes in this PR that are not enumerated in the PR description, could you add them so I can make sure they aren't lost? Thanks!

kumavis:
heres where we're at getting these changes in master (may have missed something)

- [ ] src/query/workerQueue
  - [x] cb -> async
  - [ ] uses Iterators instead of an async/queue
  - [x] stops any running workers if an error is thrown
  - [x] has a mechanism to stop a query once the required number of results have been fetched
  - [ ] ensure peers are unique by peer id (before they were only unique by Object reference)
- [ ] providers
  - [ ] cb -> async
  - [ ] now ensures that `addProvider()`, `getProviders()` and `_cleanup()` are synchronized (using a p-queue)
  - [ ] now cleans up any expired entries (previously expired entries would only be cleaned up if all peer ids for a given CID had expired)
  - [ ] add items to the cache any time we go to the datastore (previously we would cache on `addProvider()` but not on `getProviders()`)
- [ ] random-walk
  - [ ] cb -> async
  - [x] simplify scheduling mechanism
  - [x] fix a bug where we [check a non-existent variable on start](https://github.com/libp2p/js-libp2p-kad-dht/blob/master/src/random-walk.js#L33)
- [ ] dht/index
  - [ ] cb -> async
  - [ ] getMany()
    - [x] Fix bug where requesting a single value would fail if the value was not found in the local datastore (ie before it wouldn't go out to the DHT to look for it. Now it does)
    - [ ] async Iterator interface
  - [x] stop()
    - [x] Ensure all running queries are stopped when the DHT is stopped
- [ ] Move all pull-streams into a single class so that we can easily replace with Iterators when the `Connection` interface has been updated
  - [ ] src/rpc
  - [ ] src/providers
  - [ ] src/network

vasco-santos:
@kumavis @dirkmc does it make sense to close this PR and work from the current state of things?

We already have a lot of work done here, as well as the current status check list from @kumavis to iterate. I hope to get the `interface-connection` work finished soon, so that we can avance with this

PR #85, opened 2019-03-01
fix: put query for closest peers

vasco-santos:
When we have a big number of peers, the query does not stop and put hangs, also hitting heavily the CPU. Moreover, we will get a really big set of peers from the `closestPeers` and trying to put to all of them also cause problems, as some end up failing, which causes the put operation to fail.

In a long run, we have to refactor the `closestPeers` and `query` logic to be more fault tolerant. We should stream the query as we would be able to stop the query at a certain point (should have an abort logic). I will create an issue for this afterward.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -217,7 +217,7 @@ class KadDHT extends EventEmitter {
       (cb) => utils.createPutRecord(key, value, cb),
       (rec, cb) => waterfall([
         (cb) => this._putLocal(key, rec, cb),
-        (cb) => this.getClosestPeers(key, cb),
+        (cb) => this.getClosestPeers(key, { shallow: true }, cb),
         (peers, cb) => {
           // Ensure we have a default `minPeers`
           options.minPeers = options.minPeers || peers.length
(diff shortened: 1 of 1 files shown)


src/index.js
```diff
@@ -387,11 +387,18 @@ class KadDHT extends EventEmitter {
    * Kademlia 'node lookup' operation.
    *
    * @param {Buffer} key
+   * @param {boolean} isOneIteration query with one iteration
    * @param {function(Error, Array<PeerId>)} callback
    * @returns {void}
    */
-  getClosestPeers (key, callback) {
+  getClosestPeers (key, isOneIteration, callback) {
```
jacobheun: isOneIteration should be an options object for better future proofing.

jacobheun: We'll likely refactor this away anyway, but perhaps it might be good to use a more generic term like `shallow`, as it's possible we may use that later. The name change isn't a blocker though.

vasco-santos: agree, I have updated!

vasco-santos:
I will open the issue for revisiting this logic!

PR #87, opened 2019-03-06
fix: not emit peer discovery event

vasco-santos:
`kad-dht` does not need to emit the `peer-discovery` event since it already dials to the peers.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,6 +1,5 @@
 'use strict'
 
-const { EventEmitter } = require('events')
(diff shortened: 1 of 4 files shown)


vasco-santos:
coverage just decreased because I removed tested code

alanshaw:
I think anyone using libp2p will expect to receive this event for ALL peers that are discovered. It's unexpected to have some discovered peers emitted but not others.

vasco-santos:
Not needed anymore!

PR #93, opened 2019-03-28
Ensure queries stop after error or success

dirkmc:
Currently queries continue running
- after there was a timeout
- after a match is found on one disjoint path (eg in find peer)

This PR changes the query function to return either `pathComplete` or `queryComplete`, where `queryComplete` indicates that the whole query should stop (not just this particular disjoint path)

It also ensures that queries are stopped after a timeout or any other error.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -359,7 +359,7 @@ class KadDHT extends EventEmitter {
 
                 // enough is enough
                 if (pathVals.length >= pathSize) {
-                  res.success = true
+                  res.pathComplete = true
                 }
 
                 cb(null, res)
@@ -368,7 +368,12 @@ class KadDHT extends EventEmitter {
           })
 
           // run our query
-          timeout((cb) => query.run(rtp, cb), options.timeout)(cb)
+          timeout((_cb) => {
+            query.run(rtp, _cb)
+          }, options.timeout)((err, res) => {
+            query.stop()
+            cb(err, res)
+          })
         }
       ], (err) => {
         // combine vals from each path
@@ -419,7 +424,7 @@ class KadDHT extends EventEmitter {
             (closer, cb) => {
               cb(null, {
                 closerPeers: closer,
-                success: options.shallow ? true : undefined
+                pathComplete: options.shallow ? true : undefined
               })
             }
           ], callback)
@@ -636,7 +641,7 @@ class KadDHT extends EventEmitter {
                   if (match) {
                     return cb(null, {
                       peer: match,
-                      success: true
+                      queryComplete: true
                     })
                   }
 
@@ -648,9 +653,12 @@ class KadDHT extends EventEmitter {
             }
           })
 
-          timeout((cb) => {
-            query.run(peers, cb)
-          }, options.timeout)(cb)
+          timeout((_cb) => {
+            query.run(peers, _cb)
+          }, options.timeout)((err, res) => {
+            query.stop()
+            cb(err, res)
+          })
         },
         (result, cb) => {
           let success = false
--- a/src/query.js
+++ b/src/query.js
@@ -85,6 +85,8 @@ class Query {
       }
     })
 
+    // Create a manager to keep track of the worker queue for each path
+    this.workerManager = new WorkerManager()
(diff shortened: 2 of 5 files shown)


src/index.js
```diff
@@ -368,7 +368,12 @@ class KadDHT extends EventEmitter {
           })
 
           // run our query
-          timeout((cb) => query.run(rtp, cb), options.timeout)(cb)
+          timeout((_cb) => {
+            query.run(rtp, _cb)
+          }, options.timeout)((res, err) => {
```
jacobheun: `res` and `err` order is wrong, they're being passed to the callback in the same order so things are still functioning.

src/index.js
```diff
@@ -648,9 +653,12 @@ class KadDHT extends EventEmitter {
             }
           })
 
-          timeout((cb) => {
-            query.run(peers, cb)
-          }, options.timeout)(cb)
+          timeout((_cb) => {
+            query.run(peers, _cb)
+          }, options.timeout)((res, err) => {
```
jacobheun: same here, flip `res` and `err`

src/query.js
```diff
@@ -110,109 +118,170 @@ class Query {
...
+        }
 
-  q.drain = () => {
-    query._log('queue:drain')
-    callback()
-  }
+        // If query is complete, stop all workers
+        if (state && state.queryComplete) {
+          query._log('query:complete')
+          return this.stop()
```
jacobheun: I initially thought this was exiting without calling the callback, but saw that `stop` calls `worker.stop` which kills the queue. It might be worth updating the description of `stop` to mention the underlying queues will be killed.

src/query.js
```diff
@@ -110,109 +118,170 @@ class Query {
...
+          queryComplete: res.queryComplete
+        })
+      } else if (res.closerPeers && res.closerPeers.length > 0) {
+        each(res.closerPeers, (closer, cb) => {
+          // don't add ourselves
+          if (query.dht._isSelf(closer.id)) {
+            return cb()
+          }
+          closer = query.dht.peerBook.put(closer)
+          query.dht._peerDiscovered(closer)
```
jacobheun: Let's not change anything for this PR, but I think we should look at centralizing `peerBook.put` calls. Ideally I think those should all role up to libp2p via the `peer` event and it should handle it, but perhaps there are reasons for the dht to be able to do that. At the very least I think `_peerDiscovered` would be a potential spot for that change.

test/query.spec.js
```diff
@@ -69,6 +70,30 @@ describe('Query', () => {
...
+      }
+      cb(null, {
+        closerPeers: [peerInfos[2]]
+      })
+    }
+
+    const q = new Query(dht, peer.id.id, () => query)
+    q.run([peerInfos[1].id], (err, res) => {
+      expect(err).not.to.exist()
+      expect(res.finalSet.size).to.eql(2)
```
jacobheun: Can we add a comment about why this is 2 and/or verify the contents of `finalSet`? It would be helpful to see and test exactly what we expect the final set to contain. The comments you added in the other tests are super helpful.

src/query.js
jacobheun: Issue added at https://github.com/libp2p/js-libp2p/issues/345. Resolving this convo.

dirkmc: Makes sense to me 👍

src/index.js
dirkmc: Oops, nice catch :)

src/query.js
dirkmc: 👍 I added some comments to make it less confusing. The code in this class is a little hairy. In the [async-await PR](https://github.com/libp2p/js-libp2p-kad-dht/pull/82) for this package I separated everything into different classes and cleaned it up.

src/sim-server.js
```diff
@@ -0,0 +1,42 @@
+'use strict'
```
jacobheun: Looks like this file was accidentally added?

vasco-santos:
Released on `0.14.11`

PR #95, opened 2019-04-03
Stop running queries on shutdown

dirkmc:
Make sure that any running queries are stopped when the DHT stops


dirkmc:
Good catch 👍 
I pushed a fix

dirkmc:
Makes sense - I added test cases for
- passing an empty list of peers to Query.run()
- calling Query.run() after shutdown

PR #1, opened 2017-01-02
[WIP] Actually implement this

dignifiedquire:
The dragons still live here. So beware of them. 

## Things left to do 

- [x] [external] release libp2p-record https://github.com/libp2p/js-libp2p-record/pull/1
- [x] [external] Update libp2p peer events https://github.com/libp2p/js-libp2p/issues/74
- [x] [external] Make peerbook in libp2p be the canoncial list of all seen peers https://github.com/ipfs/js-libp2p-ipfs-nodejs/pull/91
- [x] ~~[external] Add latency recording to peer-book~~ (moved to https://github.com/libp2p/js-peer-book/issues/34)
- [x] ~~[external] Add close method to interface-connection~~ (moved to https://github.com/libp2p/js-libp2p-swarm/issues/205)
- [x] [external] Integrate into libp2p https://github.com/libp2p/js-libp2p/pull/86
- [ ] [external] Integrate into js-ipfs https://github.com/ipfs/js-ipfs/pull/822
- [ ] [external] Integrate into js-ipfs-bitswap https://github.com/ipfs/js-ipfs-bitswap/pull/142
- [ ] Fix all todos
- [x] ~~Figure out if go uses lp messages or not~~, It does, need to implement it 
- [x] Implement bootstrap
- [x] Implement providers interface
- [x] Implement peer query interface
- `dht_test.go`
  - [x] get-set
  - [x] providers
  - [x] bootstrap
  - [x] layered get
  - [x] findPeer
  - [x] ~~findPeersConnectedToPeer~~ (removing, doesn't work in go and will be implemented at a later point)
  - [x] connect collision
  - [x] bad proto messages
  - [x] findPeer query
  - [x] findClosestPeers
- `ext_test.go`
  - [ ] get failures
  - [ ] not found
  - [ ] less than K responses
- [x] Enough tests for good coverage stats
  - [x] Statements   : Current: 86% - Goal: 90% fine enough :)
  - [ ] Branches     : Current: 67% - Goal: 90% - will improve at a later point
  - [x] Functions    : Current: 99% - Goal: 100% fine enough :)
  - [x] Lines        : Current: 86% - Goal: 90% fine enough :)
- [ ] Write interop tests with go
  - [ ] in js-ipfs
  - [x] encoded messages
- [x] add `start` and `stop`
- [ ] Documentation
  - [x] API
  - [ ] Overview with graph and things

PR #97, opened 2019-04-10
Limit scope of queries to k closest peers

dirkmc:
Implements https://github.com/libp2p/go-libp2p-kad-dht/issues/290

Stop querying once we have successfully queried >= K peers, and the only remaining peers waiting to be queried are further from the key than those K peers.

I also fixed a bug where we would return peers that we hadn't successfully connected to and refactored the Query class.

### DIFF
--- a/src/query.js
+++ b/src/query.js
@@ -506,7 +506,7 @@ class WorkerQueue {
    */
   processNext (peer, cb) {
     if (!this.running) {
-      return
+      return cb()
     }
 
     // The paths must be disjoint, meaning that no two paths in the Query may
@@ -518,7 +518,7 @@ class WorkerQueue {
     // Check if we've queried enough peers already
     this.run.continueQuerying((err, continueQuerying) => {
       if (!this.running) {
-        return
+        return cb()
       }
 
       if (err) {
@@ -527,7 +527,7 @@ class WorkerQueue {
 
       // If we've queried enough peers, bail out
       if (!continueQuerying) {
-        return
+        return cb()
       }
 
       // Check if another path has queried this peer in the mean time
@@ -551,17 +551,18 @@ class WorkerQueue {
 
         // If query is complete, stop all workers.
         // Note: run.stop() calls stop() on all the workers, which kills the
-        // queue and calls callback(), so we don't need to call cb() here
+        // queue and calls callbackFn()
         if (state && state.queryComplete) {
           this.log('query:complete')
-          return this.run.stop()
+          this.run.stop()
+          return cb()
         }
 
         // If path is complete, just stop this worker.
-        // Note: this.stop() kills the queue and calls callback() so we don't
-        // need to call cb() here
+        // Note: this.stop() kills the queue and calls callbackFn()
         if (state && state.pathComplete) {
-          return this.stop()
+          this.stop()
+          return cb()
         }
 
         // Otherwise, process next peer
--- a/src/query.js
+++ b/src/query.js
@@ -478,11 +478,20 @@ class WorkerQueue {
   }
 
   /**
-   * Add peers to the queue until there are enough to satisfy the concurrency.
+   * Add peers to the worker queue until there are enough to satisfy the
+   * worker queue concurrency.
+   * Note that we don't want to take any more than those required to satisfy
+   * concurrency from the peers-to-query queue, because we always want to
+   * query the closest peers to the key first, and new peers are continously
+   * being added to the peers-to-query queue.
    */
   fill () {
     this.log('queue:fill')
-    while (this.queue.length() < this.concurrency &&
+
+    // Note:
+    // - queue.running(): number of items that are currently running
+    // - queue.length(): the number of items that are waiting to be run
+    while (this.queue.running() + this.queue.length() < this.concurrency &&
            this.path.peersToQuery.length > 0) {
       this.queue.push(this.path.peersToQuery.dequeue())
     }
--- a/src/query.js
+++ b/src/query.js
@@ -1,5 +1,6 @@
 'use strict'
 
+const EventEmitter = require('events')
 const waterfall = require('async/waterfall')
 const each = require('async/each')
 const queue = require('async/queue')
@@ -7,6 +8,7 @@ const mh = require('multihashes')
 
 const c = require('./constants')
 const PeerQueue = require('./peer-queue')
+const PeerDistanceList = require('./peer-distance-list')
 const utils = require('./utils')
 
 /**
@@ -42,8 +44,12 @@ class Query {
     this.dht = dht
     this.key = key
     this.makePath = makePath
-    this.concurrency = c.ALPHA
     this._log = utils.logger(this.dht.peerInfo.id, 'query:' + mh.toB58String(key))
+
+    this.running = false
+
+    this._onStart = this._onStart.bind(this)
+    this._onComplete = this._onComplete.bind(this)
   }
 
   /**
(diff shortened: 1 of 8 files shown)


dirkmc:
There are four places we perform queries:
- getClosestPeers()
- findPeer()
- findProviders()
- getMany()

I think it makes sense to implement this behaviour for the first two, but I'm not sure about findProviders() and getMany(). @jhiesey @vasco-santos @jacobheun I'd like to hear your thoughts.

jacobheun:
I think the same logic applies for `findProviders()` and `getMany()`. Is your concern around these two not going deep enough into the network to find potentially closer peers?

dirkmc:
findProviders() and getMany() use disjoint paths whereas getClosestPeers() and findPeer() do not, so I wanted to consider what effect that might have.

For example when running getMany(), we first work out the number of disjoint paths, which will be equal to alpha (ie 3). Results are divided among the paths, such that each path can return at most `Math.ceil(valueCount / numPaths)`. So if we attempt to fetch 16 values with getMany() (as IPNS does by default) the number of values per path will be `Math.ceil(16 / 3) = 6`. If K=20, the value is stored at 20 peers. Consider a topology where there are 8 values along one path, 9 values along the second path and 3 along the third path. In the best case we will return 6, 6 and 3 = 15 values (because the maximum per path is 6).

Limiting the scope of the query makes it run faster, because we examine less nodes, but it also makes it less likely that we will discover all possible paths to the values, meaning getMany() is more likely to fail.

jacobheun:
> Limiting the scope of the query makes it run faster, because we examine less nodes, but it also makes it less likely that we will discover all possible paths to the values, meaning getMany() is more likely to fail.

It sounds like this is where properly streaming responses would make it easier to achieve both. Until we add the streaming responses, we'll likely have to have the limits on those two be higher, because that seems likely to fail a lot. The math is not in IPNS's favor.

dirkmc:
Streaming responses should help. For now we could also decrease the required number of values for IPNS to be less than 16. There can only be one source crypto key for IPNS records, which in practice usually means only one source node, so IPNS records will almost always all have the same value for the same key (they would only differ when an IPNS publish hasn't reached all the nodes that a previous IPNS publish reached).

I implemented a streaming version of the DHT in this PR: https://github.com/libp2p/js-libp2p-kad-dht/pull/82. It's waiting on async/await PRs for a few dependencies to be approved.

test/kad-dht.spec.js
```diff
@@ -686,54 +688,89 @@ describe('KadDHT', () => {
...
             kadUtils.sortClosestPeers(out, rtval, (err, got) => {
               expect(err).to.not.exist()
+
+              // TODO:
+              // With this change this expectation fails intermittently,
+              // because getClosestPeers() doesn't always find the closest 20
+              // peers (eg sometimes it will find 18 of the closest plus a
+              // couple of others that are not quite the closest).
+              // I'm not sure what the best approach to testing should be here?
               expect(countDiffPeers(exp, got)).to.eql(0)
```
dirkmc: I'm not sure what the best approach to testing should be here?

jacobheun:
```js
// Make connections from each of the first 20 nodes to 16 of the
// remaining 80 nodes at random
```

Perhaps we should specify the topology for this so we can get deterministic results? If we pick nodes at random we're going to get random results. If we want a random dht network test, we can't have hard specifications because it's not a deterministic environment.  We'd have to do things like:

```js
// If results are greater than 20% different from the ideal set
// in a network of this size, there is problem and it needs to be fixed.
// If this test is failing DO NOT increase `.at.most` beyond 20% (4), there's
// a problem and it needs to be fixed.
expect(countDiffPeers(exp, got)).to.be.at.most(4)
```

I do think a better test is to specify the Topology. We should be able to get deterministic results that way correct? Is there any level of randomness that could affect that?

dirkmc: That's true. An issue with this test is that it's not a very realistic topology. In reality, nodes close to a particular key should know about most of the other nodes close to the key. I will adjust the test to reflect that.

vasco-santos: We should do a deterministic test like jacob suggest, followed by a more real test, where we do not assume number like in this test.

vasco-santos:
I agree with @dirkmc and starting by targetting `getClosestPeers()` and `findPeer()`!

After that, once we introduce the stream we figure out what limits we can add to the remaining two.

dirkmc:
I think I've found a reason that the find peer test is often failing.

I noticed that [by default](https://github.com/libp2p/js-libp2p-kad-dht/blob/master/src/index.js#L83) we [limit the number of peers](https://github.com/libp2p/js-libp2p-kad-dht/blob/master/src/private.js#L37) returned from the FIND_NODE RPC to 6, whereas the Kademlia spec ([and go-libp2p-kad-dht](https://github.com/libp2p/go-libp2p-kad-dht/blob/master/handlers.go#L23)) specify that it should be KValue (20 by default).

This will cause a problem in networks where many nodes have accurate knowledge of the closest peers to a particular key (eg if that key is requested frequently). For example in the following scenario:
- alpha (3) peers are chosen to initiate a getClosestPeers() request
- each of the 3 peers returns the 6 closest peers to the key
- each of those 6 closest peers return each other
- the total number of peers queried will end up being low, potentially as low as 10

This will cause a problem for some DHT operations, for example when doing a PUT to the DHT, this will cause a problem because it needs to store the PUT value on at least KValue (20) nodes.

Does anyone know why this limit of 6 was chosen?

Here's what the [Kademlia spec](http://www.scs.stanford.edu/~dm/home/papers/kpos.pdf) says:
> FIND NODE takes a 160-bit ID as an argument. The recipient of the RPC returns <IP address, UDP port, Node ID> triples for the k nodes it knows about closest to the target ID. These triples can come from a single k-bucket, or they may come from multiple k-buckets if the closest k-bucket is not full. In any case, the RPC recipient must return k items (unless there are fewer than k nodes in all its k-buckets combined, in which case it returns every node it knows about).

dirkmc:
I fixed a bug in the queueing logic, and changed the test so that the topology is more realistic (nodes know about peers close to them), so the tests now pass consistently

jacobheun:
> Does anyone know why this limit of 6 was chosen?

No, and it looks like it's been there since day 1, https://github.com/libp2p/js-libp2p-kad-dht/pull/1. We should use what's in the spec unless we have clear, documented reasoning for deviating. 


It looks like the bundlesize is failing, with the added code this is understandable. We should bump https://github.com/libp2p/js-libp2p-kad-dht/blob/v0.14.12/.aegir.js#L2 to `192kb`.

vasco-santos:
@dirkmc did you check if in go land, they also use `6` ? If they do not, I am totally in favor of following the spec

jacobheun:
> whereas the Kademlia spec (and go-libp2p-kad-dht) specify that it should be KValue (20 by default)

While go is following the spec on this item, we really should avoid doing things here based on the go implementation. The go implementation has known issues and regularly references needing to follow the Kademlia spec properly. I think we should focus energy in making sure we're abiding by original Kademlia spec and pushing the libp2p spec for it, https://github.com/libp2p/specs/pull/108, forward, rather than attempting to mimic the go implementation. 

As long as we're not breaking interop, and we document where and why we're deviating from go (and informing them if they're not aware), I think that's the approach we should be taking.

test/query.spec.js
```diff
@@ -570,6 +578,99 @@ describe('Query', () => {
...
+          const closerPeers = closerIndexes.map(j => peerIdToInfo(sorted[j]))
+          setTimeout(() => cb(null, { closerPeers }))
+        }
+
+        const q = new Query(dht, peerInfos[0].id.id, () => query)
+        q.run(initial, (err, res) => {
+          expect(err).to.not.exist()
+
+          // Should query 20 peers and then stop
+          expect(visited.length).to.be.gt(20)
```
jacobheun: Based on the current Topology this should be at least 25 right? Assuming I counted correctly we should have all unique values from below. If that's correct, I think we should actually validate that visited actually contains all of those values.

```js
          10: [12, 20, 22, 24, 26, 28],
          16: [14, 18, 20, 22, 24, 26],
          18: [4, 6, 8, 12, 14, 16],

          26: [24, 28, 30, 38],
          30: [14, 28],
          38: [2],

          // Should zoom out from this point, until it has 20 peers
          2: [13],
          13: [15],
          15: [17],

          // Right before we get to 20 peers, it finds some new peers that are
          // closer than some of the ones it has already queried
          17: [1, 3, 5, 11],
          1: [7, 9],
          9: [19],
```

Other than that, I think this test is a big improvement 👍

dirkmc: Makes sense 👍 
Fixed

dirkmc:
I changed the number of closer peers to be returned to K (@vasco-santos the go DHT does use the k value also: https://github.com/libp2p/go-libp2p-kad-dht/blob/master/handlers.go#L23)

src/query.js
```diff
@@ -58,266 +64,565 @@ class Query {
...
+        return
+      }
+
+      if (err) {
+        return cb(err)
+      }
+
+      // If we've queried enough peers, bail out
+      if (!continueQuerying) {
+        return
```
jacobheun: In a few places in this function, such as here, it's possible that `cb` never gets called. Is this intended? Will this potentially lead to stalled/stuck queues?


Other than this, things look good!

dirkmc: It only ever returns without calling `cb` after the workers have all been stopped, but I think it's less confusing if we call `cb` in that situation anyway, so I changed it to do so.

dirkmc:
@jacobheun do you think we should just enable this for getClosestPeers() and findPeer(), or for all four queries including findProviders() and getMany()?

jacobheun:
Perhaps we should start with `getClosestPeers()` and `findPeer()`, and then make a determination about the other two? Outlining the pros and cons for them would be helpful.

It might be valuable to have the logic in `findProviders()` moreso than `getMany()`, but it we can't yet stream, having outright failures is not good. 

One of the current unfortunate things about `findProviders()` having that threshold, is that getting new content can be potentially impossible to get. For example, in ipfs-bitswap, when I try to get new content you've created, if you're the only provider when I search, I'll never get the content because 1 is less than my minimum threshold. It will eventually timeout because I can't find 20 peers. For DHT performance in IPFS we have to fix that.

Maybe it's worth adding findProviders until we can properly support streaming?

vasco-santos:
@jacobheun totally agree that we should follow the kademlia spec first and then try to iterate on improving it. In that time, we may check the go implementation looking for possible optimizations they did to the paper. We just need to make this visible for everyone being on the same page.

Regarding the 4 functions that perform queries, I think we should do this by steps. I would go with these two now. However, I agree that it would be worth exploring this for `findProviders` as well. My proposal is to get the first two merged and then work on adding the `findProviders` in a separate PR. That way, we can easily test the differences and revert `findProviders` if needed. What do you think? @dirkmc @jacobheun

dirkmc:
Thinking about it a bit more, we may need to decide for each case in which we call any of those four functions. For example when doing a DHT PUT we call `getClosestPeers()` and store the value at each of those peers. We want to be sure that those really are the closest peers or it will decrease the chance of `getMany()` finding enough values.

vasco-santos:
@dirkmc do you think it is worth putting together a list of all the possible cases and discuss for each of them which is the best solution. We should also get @jhiesey involved on these decisions

dirkmc:
Yes I think that makes sense, I will put together a list

dirkmc:
The paths through the code that create a `Query` are:
- getMany()
- findProviders()
- findPeer()
  - As a public API method
  - Called by random walk to discover peers
- getClosestPeers()
  - As a public API method
  - Called by provide() to add self as a provider for a key
  - Called by put() to find nodes on which to store a key / value in the DHT

jacobheun:
So `getClosestPeers()` belongs only to write actions to the network. This makes sense because we want to be accurate when we're storing values. The downside of this is that `getClosestPeers` should, by design, be "slow", moreso for newer peers with sparse routing tables.

For the fetching functions `getMany`, `findPeer`, `findProviders` we should be as quick as possible. As mentioned, I think this may warrant reducing the tolerance on number of peers required for a success. Finding 16 records when there are potentially only 20 nodes who have values feels too high to me, especially if the query should be relatively quick. With a network that is still in flux, with so many new nodes coming and going, I think lowering the threshold at least initially is a reasonable approach.

For eventually getting the tolerance back up to a higher level (16), there's discussion about adopting sloppiness at https://github.com/libp2p/specs/issues/163. While it's not the same concept of sloppiness, I was thinking about how being sloppy with the number of writes, combined with eventual accuracy, might be helpful in alleviating some of this. For example, we discussed concerns about spending enough time finding closest peers when performing a `PUT`, so maybe exiting early isn't ideal, there could be closer peers. If we operated "sloppily" we could perform a `PUT` with the closest kValue (20) we have when we would normally end the query, and then continue searching looking for eventual accuracy. Once the query was "done", we could `PUT` to any closer peers we hadn't already. This breaks the Spec a bit because we're potentially writing to more than kValue peers, but I think there is merit in the fault tolerance it provides to the network.

dirkmc:
That's an interesting idea. I wonder in practice how often we won't find the actual closest 20 peers. Maybe it's worth merging this in and trying it out in the real network?

jacobheun:
> Maybe it's worth merging this in and trying it out in the real network?

Yes, I think getting this out so we can test with it and iterate is a good path forward.

vasco-santos:
Thanks for the list @dirkmc  

I liked your proposal @jacobheun 

I will merge this as it looks good and we need to test it, but I believe we will want to iterate on this. 
There is important information that we will need for the next iterations, so I will create an issue for tracking the good ideas from this PR and we can keep discussing there.

Issue #100, opened 2019-04-15
Proposal: add trigger to randomWalk in addition to interval


jacobheun:
Currently `randomWalk`, when enabled, triggers on a given interval (current default of 5mins). When a node is starting, if it uses `randomWalk` as a peer discovery mechanism, it can take 5 minutes to crawl the dht for new peers. Assuming you launch your node and take no dht actions (as these should result in discovered peers), this can take a while to gain a meaningful number of peers and contribute to the network.

Adding a trigger could enable us to find peers more quickly, and leave the randomWalk interval higher. My thought is that without a trigger, it's more likely that the randomWalk interval will be reduced to help with initial discovery which feels wasteful in terms of queries to the dht. If I leverage another discover mechanism like `libp2p-bootstrap`, I could configure my trigger for something like `10s` and leave the randomWalk interval at the default 5mins. By 10s I should be connected to several of the bootstrap nodes. Performing the randomWalk should result in many discovered peers, which in turn will lead to more connected peers, especially with libp2p auto dial on. The more quickly we're able to "bootstrap" into a good portion of the network, we'll also reduce our dependency on the Bootstrap nodes, which is beneficial for their health as well.

cc: @alanshaw

dirkmc:
I like this idea :+1:

Could you explain how the trigger mechansm would work? Is it a one-time event that causes random walk to start?

jacobheun:
> Is it a one-time event that causes random walk to start?

Yes, I guess `delay` is a better term. Currently, when we start randomWalk it waits for the interval period to occur. If we change the randomWalk to **also** execute a walk as soon as it starts, we can use the `delay` to start the service. This way once the `delay` as occurred we'll immediately execute the walk, and then run it on `interval` after that.

So a randomWalk config of:
```json
enabled: true,
interval: 300e3,
delay: 10e3,
...
```

would execute a walk at `10s, 5min 10s, 10m 10s, ...,` once the dht has started.

dirkmc:
Can we just start the DHT with `{ randomWalk: { enabled: false }}`, and then call `dht.randomWalk.start()` once we are connected to enough bootstrap nodes?

jacobheun:
Potentially, but this gets complicated because we don't do any tagging of peers yet. As the number of discovery services grow we need to know which peers are Bootstrap nodes, or more generally, which peers have DHTs. We do have the support to check the protocols a peer is running, so we could check who has DHTs once we have a muxed connection to them.

[...]

PR #101, opened 2019-04-16
feat: add delay support to random walk

jacobheun:
resolves https://github.com/libp2p/js-libp2p-kad-dht/issues/100

This adds support for `delay` in the random walk config. This allows random walk to be run more quickly after startup, while maintaining a longer interval.

I ran this against js-ipfs master with a randomWalk config (all options aside from enabled are default):
```js
randomWalk: {
  enabled: true,
  interval: 300e3,
  delay: 10e3,
  timeout: 10e3
}
```

With an idle node, I had 100+ connected peers in under a minute. The connected peers eventually stabilized around the max peers of ~300.

**logging**: I also updated the randomConfig logging to make it easier to target for debugging. Instead of `DEBUG=libp2p:dht` and sifting through the logs, you can now have `DEBUG=libp2p:dht:random-walk*` and follow only random walk logs, and error logs.

**Note**: libp2p fix https://github.com/libp2p/js-libp2p/pull/359 will need to be released before `delay` can be configured by js-ipfs or other projects.

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -44,5 +44,6 @@ exports.defaultRandomWalk = {
   enabled: true,
   queriesPerPeriod: 1,
   interval: 5 * minute,
-  timeout: 10 * second
+  timeout: 10 * second,
+  delay: 10 * second
 }
--- a/src/index.js
+++ b/src/index.js
@@ -24,7 +24,6 @@ const Message = require('./message')
 const RandomWalk = require('./random-walk')
 const QueryManager = require('./query-manager')
 const assert = require('assert')
-const mergeOptions = require('merge-options')
 
 /**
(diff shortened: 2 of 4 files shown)


vasco-santos:
@dirkmc I agree that it would be great to add it, as the results obtained from an ongoing query will not be relevant. I think we can add it in a new PR, as this is looking good to go. What do you guys think?

dirkmc:
Agreed - I'm find with it being done in a new PR

vasco-santos:
Created #102 for tracking that. It should be an easy addition to the random-walk codebase

jacobheun:
Yes it's a good idea, and actually I think the existing randomWalk query logic is wrong right now. I'll look at a fix for it with the timeout update.

PR #104, opened 2019-04-18
fix: random walk

jacobheun:
Random Walk had some bugs, this fixes them: 
* If you ever specified more than 1 query, it would never run them. (You really shouldn;t need more than 1, but still not a good bug)
* We timed out the walk, but `findPeer` wasn't given the walks timeout. So if you specified 10seconds as your walk timeout, the `findPeer` logic would keep running until its default timeout kicked in (60 seconds)

I added a suite of tests just for RandomWalk to test the failures I was seeing.

This also adds support for AbortController in random walk and passes it to findPeer, **however**, findPeer currently isn't listening to the abort signal so it won't actually stop the query. My thought is to wait until we finish #82 and then add the abort logic to all of the query methods. Once we do, random walk will immediately benefit because it already passes the signal in. @dirkmc @vasco-santos let me know what you think, and I can update accordingly.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -580,7 +580,7 @@ class KadDHT extends EventEmitter {
    * @param {PeerId} id
    * @param {Object} options - findPeer options
    * @param {number} options.timeout - how long the query should maximally run, in milliseconds (default: 60000)
-   * @param {EventEmitter} options.signal - an `AbortController` signal
+   * @param {AbortControllerSignal} options.signal - an `AbortController` signal
    * @param {function(Error, PeerInfo)} callback
    * @returns {void}
    */
--- a/src/index.js
+++ b/src/index.js
@@ -580,7 +580,6 @@ class KadDHT extends EventEmitter {
    * @param {PeerId} id
    * @param {Object} options - findPeer options
    * @param {number} options.timeout - how long the query should maximally run, in milliseconds (default: 60000)
-   * @param {AbortControllerSignal} options.signal - an `AbortController` signal
    * @param {function(Error, PeerInfo)} callback
    * @returns {void}
    */
--- a/src/index.js
+++ b/src/index.js
@@ -173,11 +173,10 @@ class KadDHT extends EventEmitter {
    */
   stop (callback) {
     this._running = false
-    this.randomWalk.stop(() => { // guarantee that random walk is stopped if it was started
-      this.providers.stop()
-      this.network.stop(callback)
-    })
+    this.randomWalk.stop()
+    this.providers.stop()
     this._queryManager.stop()
+    this.network.stop(callback)
   }
 
   /**
@@ -581,6 +580,7 @@ class KadDHT extends EventEmitter {
    * @param {PeerId} id
    * @param {Object} options - findPeer options
    * @param {number} options.timeout - how long the query should maximally run, in milliseconds (default: 60000)
+   * @param {EventEmitter} options.signal - an `AbortController` signal
    * @param {function(Error, PeerInfo)} callback
    * @returns {void}
    */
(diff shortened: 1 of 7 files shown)


src/index.js
```diff
@@ -581,6 +580,7 @@ class KadDHT extends EventEmitter {
    * @param {PeerId} id
    * @param {Object} options - findPeer options
    * @param {number} options.timeout - how long the query should maximally run, in milliseconds (default: 60000)
+   * @param {AbortControllerSignal} options.signal - an `AbortController` signal
```
dirkmc: Does the DHT constructor need an AbortControllerSignal parameter?

src/random-walk.js
```diff
@@ -52,23 +52,18 @@ class RandomWalk {
           runningHandle._timeoutId = null
```
dirkmc: It seems like we're no longer using `runningHandle._onCancel`. Maybe the code could be simplified by replacing `this.runningHandle` with `this.timeoutId`, and making `runPeriodically` a class method?

src/index.js
jacobheun: I think I will remove this from the findPeer method until it actually supports it.

> I'm not sure I'd include the AbortController logic until we move to an async/await API

Do you think we should just remove it from randomWalk altogether, or leave what's in there? When we do the update in async/await if `findPeer` takes `options.signal` as a parameter, we won't have to do any work there to hook up randomWalk. Adding an abort test to findPeer will verify it's working, since randomWalk is validating it's passing the param onward.

src/random-walk.js
jacobheun: Good call, I will review the code there.

src/random-walk.js
```diff
@@ -113,39 +108,53 @@ class RandomWalk {
...
+      callback(err)
     })
   }
 
   /**
    * The query run during a random walk request.
    *
    * @param {PeerId} id
+   * @param {object} options
+   * @param {number} options.timeout
```
vasco-santos: Add `options.signal`

jacobheun:
I've removed the runningHandler logic and simplified the code. I also bumped up the default timeout to 30s from 10s. Since we are passing on the timeout to findPeer, and it actually times the query out, giving it more time to run will improve the effectiveness.

I also disabled random walk in the TestDHT util by default to keep the tests isolated better. We're managing connects as needed in the tests, having random walk running will just cause more test chaos. 

I added a comment to the `RandomWalk._query` function about the AbortController not yet being properly supported and removed the docs for `options.signal` from findPeer.

test/kad-dht.spec.js
```diff
@@ -73,7 +73,7 @@ function connect (a, b, callback) {
 
 function bootstrap (dhts) {
   dhts.forEach((dht) => {
-    dht.randomWalk._walk(1, 1000, () => {})
+    dht.randomWalk._walk(1, 10000, () => {})
```
jacobheun: Now that timeout is actually being passed to `findPeer` I bumped this to 10seconds so the random walk doesn't get killed. This was causing test failures because routing tables weren't doing enough connecting when it was depending on the walk.

src/random-walk.js
```diff
@@ -41,64 +40,46 @@ class RandomWalk {
...
+  /**
+   * Run function `walk` on every `interval` ms
+   * @param {function(callback)} walk The function to execute on `interval`
+   * @param {number} interval The interval to run on in ms
+   *
+   * @private
+   */
+  _runPeriodically (walk, interval) {
+    this._timeoutId = setTimeout(() => {
+      this._timeoutId = null
```
dirkmc: I think setting this._timeoutId to null here means that start() can be called twice?

jacobheun: Yes, good catch. I updated the double start test to better validate the internals which exposed the double start was indeed happening. Nullifying the timeoutId is now only done on stop, since it only needs to happen there.

PR #115, opened 2019-05-10
Bug Fix: rename query.run object to not overwrite query.run method

### LINKED ISSUE #114: Query - run is overwritten
Query has a method called `run`. Query also assigns to `this.run` which would shadow the method. Is this a mistake or intentional? If its intentional its surprising and should be heavily commented.

https://github.com/libp2p/js-libp2p-kad-dht/blob/ddf80fed9ca7e8b1ec096f3160535aa3fdb45b50/src/query/index.js#L56-L67

kumavis:
Fixes #114


vasco-santos:
The changes were integrated in #116

PR #119, opened 2019-05-18
refactor(async): _findNProviders query function

kumavis:
last one in `private.js`

let me know if something is preferred over `callbackify`

related to #82


kumavis:
@vasco-santos done

PR #120, opened 2019-05-18
refactor(async): Query

kumavis:
non-breaking async refactor for
- `Query`
- `Run`
- `Path` 
- `PeerQueue`

this is intended to *not* introduce any changes in behavior

related #82

### DIFF
--- a/src/peer-queue.js
+++ b/src/peer-queue.js
@@ -4,7 +4,6 @@ const Heap = require('heap')
 const distance = require('xor-distance')
 const debug = require('debug')
 const promisify = require('promisify-es6')
-const promiseToCallback = require('promise-to-callback')
 
 const utils = require('./utils')
 
--- a/src/peer-queue.js
+++ b/src/peer-queue.js
@@ -66,10 +66,10 @@ class PeerQueue {
    * @returns {void}
    */
   enqueue (id, callback) {
-    promiseToCallback(this._enqueueAsync(id))(callback)
+    promiseToCallback(this.enqueueAsync(id))(callback)
   }
 
-  async _enqueueAsync (id) {
+  async enqueueAsync (id) {
     log('enqueue %s', id.toB58String())
     const key = await promisify(cb => utils.convertPeerId(id, cb))()
 
--- a/src/peer-queue.js
+++ b/src/peer-queue.js
@@ -3,6 +3,8 @@
 const Heap = require('heap')
 const distance = require('xor-distance')
 const debug = require('debug')
+const promisify = require('promisify-es6')
+const promiseToCallback = require('promise-to-callback')
 
 const utils = require('./utils')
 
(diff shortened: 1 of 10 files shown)


kumavis:
**not included in this PR**, I started on `WorkerQueue` but found that wrapping async implementation triggered some test failures. Looked like some deterministic race conditions in
`this.run.continueQuerying` vs `run._continueQueryingAsync`

one failure was [here](https://github.com/libp2p/js-libp2p-kad-dht/blob/ddf80fed9ca7e8b1ec096f3160535aa3fdb45b50/test/query/index.spec.js#L163) where i was seeing 9 instead of 3.

jacobheun:
Since Query and the other internal classes aren't exported, wouldn't it be better to just make all of that async directly? It looks like a lot of the methods have been promisified and/or have async counterparts. Ideally we'd only promisify other libs that have not been migrated, and callbackify the dht public api. Since this is only for the Query migration I understand that it's internally used methods would need to be callbackified, but the other classes should be able to be fully async.

kumavis:
I wasn't sure if the other classes were considered public APIs or not, with respect to breaking changes. Thanks for the clarification.

I'll take a look at how much of the existing callback APIs are still in use in other parts of the package.

kumavis:
This PR takes us across the bundlesize, but note that I did not introduce any additional deps here

kumavis:
@jacobheun ready to go

- [x] remove callback based methods in private classes
  - [x] peer-queue
    - [x] cb methods not used elsewhere
    - [x] cb methods not used by tests
    - [x] cb methods removed 
  - [x] query/index - cb methods still used
    - [x] cb methods not used elsewhere
    - [x] cb methods not used by tests
    - [X] cb methods removed
  - [x] query/path
    - [x] cb methods not used elsewhere
    - [x] cb methods not used by tests
    - [x] cb methods removed
  - [x] query/run
    - [x] cb methods not used elsewhere
    - [x] cb methods not used by tests
    - [x] cb methods removed
  - [x] query/workerQueue
    - [x] cb methods not used elsewhere
    - [x] cb methods not used by tests
    - [x] cb methods removed

kumavis:
ok, fixed a commit failing commitlint that was several commits back using interactive rebase. bleh

src/peer-queue.js
```diff
@@ -17,34 +18,22 @@ class PeerQueue {
...
-  static fromPeerId (id, callback) {
-    utils.convertPeerId(id, (err, key) => {
-      if (err) {
-        return callback(err)
-      }
-
-      callback(null, new PeerQueue(key))
-    })
+  static async fromPeerId (id) {
+    const key = await promisify(cb => utils.convertPeerId(id, cb))()
```
jacobheun: This should throw now right? Should we add `@throws` to the jsdocs? It will make it easier for anyone consuming to know that they should ensure they're handling it. If so, we should add this to the other methods that also throw.

kumavis: @jacobheun  Thats correct. What is the correct way to document throws for async functions?

jacobheun: Looking more through the jsdocs repo there's not really a correct way, unfortunately. Ideally we'd be able to do something like this, https://github.com/jsdoc/jsdoc/issues/1467, to be explicit. 

For now, either way we have to do some level of assumption looking at the docs. Since the function is async, a throw is just a promise rejection. So we either need to catch it, or `.catch` the promise.

I suppose it makes sense to leave it out for now until the docs can be explicit.

PR #121, opened 2019-05-18
refactor(async): WorkerQueue

kumavis:
putting up this incomplete PR for discussion on test failures

this diff https://github.com/libp2p/js-libp2p-kad-dht/commit/bce9556d41e94e69fbe0a7d51b2444abeddc9a9c triggers these test failure https://gist.github.com/kumavis/34ed55927b6749645d0d1c8406c39ef1

this line https://github.com/libp2p/js-libp2p-kad-dht/commit/bce9556d41e94e69fbe0a7d51b2444abeddc9a9c#diff-ef69a013ba3a73c2ad2672948fb7bab9R219 changes this error

```
Uncaught AssertionError: expected 9 to deeply equal 3
 ```
to
```
Uncaught AssertionError: expected 8 to deeply equal 3
 ```
from this assertion https://github.com/libp2p/js-libp2p-kad-dht/blob/bce9556d41e94e69fbe0a7d51b2444abeddc9a9c/test/query/index.spec.js#L163

includes commits from https://github.com/libp2p/js-libp2p-kad-dht/pull/120

### DIFF
--- a/src/query/workerQueue.js
+++ b/src/query/workerQueue.js
@@ -1,7 +1,8 @@
 'use strict'
 
-const each = require('async/each')
 const queue = require('async/queue')
+const promisify = require('promisify-es6')
+const promiseToCallback = require('promise-to-callback')
 
 class WorkerQueue {
   /**
@@ -20,6 +21,9 @@ class WorkerQueue {
 
     this.concurrency = this.dht.concurrency
     this.queue = this.setupQueue()
+    // a container for resolve/reject functions that will be populated
+    // when execute() is called
+    this.execution = null
   }
 
   /**
@@ -68,7 +72,11 @@ class WorkerQueue {
     this.running = false
     this.queue.kill()
     this.log('worker:stop, %d workers still running', this.run.workers.filter(w => w.running).length)
-    this.callbackFn(err)
+    if (err) {
+      this.execution.reject(err)
+    } else {
+      this.execution.resolve()
+    }
   }
 
   /**
@@ -78,9 +86,18 @@ class WorkerQueue {
    * @param {function(Error)} callback
    */
   execute (callback) {
+    promiseToCallback(this._executeAsync())(callback)
+  }
+
+  async _executeAsync () {
     this.running = true
-    this.callbackFn = callback
+    // store the promise resolution functions to be resolved at end of queue
+    this.execution = {}
+    const execPromise = new Promise((resolve, reject) => Object.assign(this.execution, { resolve, reject }))
+    // start queue
     this.fill()
+    // await completion
+    await execPromise
   }
 
   /**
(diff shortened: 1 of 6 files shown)


src/query/workerQueue.js
```diff
@@ -109,73 +126,85 @@ class WorkerQueue {
...
+      continueQuerying = await this.run._continueQueryingAsync(this)
+    } catch (err) {
+      continueQueryingError = err
+    }
 
-      if (err) {
-        return cb(err)
-      }
+    // Abort and ignore any error if we're no longer running
+    if (!this.running) {
```
jacobheun: This check is done at the start of the method, so this is redundant as I don't believe `this.run._continueQueryingAsync` will change the `running` variable.  This also means the try-catch isn't needed, since you're throwing immediately after this check if the error exists. That will help make this logic a bit cleaner.

The code
```js
if (!continueQuerying) {
  this.stop()
  return
}
```
is what's triggering the early stop in the test suite. It's a "sloppy" approach to the lookup to help paths run more quickly. This would potentially cause the difference of 8/9 if there were timeouts in the query promise resolution.

I'm wondering if the test failure is due to the queryFunc being promisified. In the test that's just a stub, I would assume it wound't have an issue, but perhaps there's some odd edge case with sinon and the arguments order.

The other thought is if continueQuerying is throwing an error, the stop code won't be called, and additional queries would continue to happen on this path.

kumavis: I agree the try-catch-check-throw is a bit unwieldy, but I was trying to maintain exact behavior as before https://github.com/libp2p/js-libp2p-kad-dht/blob/e493987976e4337a619b7e53cf469672a2987f1f/src/query/workerQueue.js#L122-L138

due to https://github.com/libp2p/js-libp2p-kad-dht/issues/123 I didn't make too much of an effort to analyze and factor out existing behavior

If you'd like I can see if its safe to remove it in this PR but its arguably sort of out of scope

kumavis: > This would potentially cause the difference of 8/9 if there were timeouts in the query promise resolution.

That makes sense but the original test expects 3

> The other thought is if continueQuerying is throwing an error, the stop code won't be called, and additional queries would continue to happen on this path.

Ah thats, interesting... should it call stop? That does sort of seem like an unexpected fatal error. Maybe out of scope for the PR, but open to fixing. Please advise.

jacobheun: Ah, right, the original callback code would also return early on error, so that's not the problem.

So, I ran the tests locally and it looks like this is just a side effect of moving to async. With the async changes the first 3 queries all end up finishing together, so when they check if they should continue querying, the have closer peers from the index 1 query. In the callback version, the queries finish in serial, so when the first query is done it sees there are no closer peers (because the 2nd query hasn't finished yet) and it stops the path.

If we add a slight delay to the second query in the tests, we get the expected behavior, and really the test should've done this to begin with. Updating the querySpy for call 1 resolves the issue:

```js
querySpy.onCall(1).callsFake((_, cb) => {
  setTimeout(() => cb(null, { closerPeers: returnPeers }), 10)
})
```

kumavis:
Thanks @jacobheun I'll add this to the other PR

PR #125, opened 2019-05-19
Feature Request - query internal data

kumavis:
This is a quick hack for getting some internal info from the queries. I used this to start building some graphs that show off the query flow.

image
image (1)
```
pink - query initiator
orange - query initial peer
yellow - introduced
green - provider
blue - uninvolved
black - uninvolved and not connected to telemetry
```

note: for layout-circle the positions are random, not based on peerId, though I should do that next

a link means someone suggested someone as a closer peer or provider, or connects query initiator to initial peers 
changing circle-layout to position by peerId will give an interesting look at how querys hone in on a result

It would be nice if possible to gather this info generically at the query layer, but here I put some extra info in the queryFn. When doing it all in the queryFn, the only thing missing was the initialPeers for the path. Though, in the end, I didn't need anything in the graph that was specific to a Path, but it could be helpful debugging and testing queries in the future.

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,6 +45,7 @@
     "cids": "~0.7.0",
     "debug": "^4.1.1",
     "err-code": "^1.1.2",
+    "fromentries": "^1.1.0",
     "hashlru": "^2.3.0",
     "heap": "~0.2.6",
     "interface-datastore": "~0.6.0",
--- a/src/private.js
+++ b/src/private.js
@@ -538,7 +538,7 @@ module.exports = (dht) => ({
     const paths = []
     const query = new Query(dht, key.buffer, (pathIndex, numPaths) => {
       // This function body runs once per disjoint path
-      const pathSize = utils.pathSize(out.length - n, numPaths)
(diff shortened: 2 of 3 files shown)


jacobheun:
@dirkmc put together a visualizer not too long ago as well, https://github.com/dirkmc/js-libp2p-kad-dht-viz. 

Being able to see what's going on in the DHT is a huge help for diagnosing issues. It would be great to have some mechansim of tracking the queries built in, even if it needed to be enabled via config (in case it adds non trivial overhead).

It would be great to see what we can take from these two implementations to accomplish this so that metrics are easily consumable by visualizers.

@vasco-santos did the introspection discussions for the libp2p visualization bring up the DHT at all?

vasco-santos:
Nice experiments @kumavis 

@jacobheun yes it did and it will be part of the first screens to be worked on 🚀

vasco-santos:
FYI we are working on [libp2p/go-libp2p#947](https://github.com/libp2p/go-libp2p/issues/947). Once we support it in js, we should be able to provide this

PR #127, opened 2019-05-29
Fix: send correct payload in ADD_PROVIDER RPC

dirkmc:
In the payload for ADD_PROVIDER we were sending remote peers their own id instead of the local node's id. The tests still passed because on the receiving end we ignore the payload and just look at the connection to get the provider peer id.

### DIFF
--- a/.aegir.js
+++ b/.aegir.js
@@ -1,4 +1,4 @@
 module.exports = {
-    bundlesize: { maxSize: '196kB' }
+    bundlesize: { maxSize: '197kB' }
 }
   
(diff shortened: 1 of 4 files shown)


dirkmc:
Note that the go implementation of the DHT *does* [reject ADD_PROVIDER messages](https://github.com/libp2p/go-libp2p-kad-dht/blob/master/handlers.go#L362) where the peer ID in the payload does not match the remote peer id, so this fix should improve provides

daviddias:
Excellent catch, @dirkmc !!

Are you adding a tiny interop test to ipfs/interop that checks IPFS providing to a go-ipfs network?

dirkmc:
That's a good idea, I will add a test once this is merged: https://github.com/ipfs/interop/pull/36

Issue #128, opened 2019-05-30
ADD_PROVIDER RPC handling


dirkmc:
In the [handler for the ADD_PROVIDER RPC](https://github.com/libp2p/js-libp2p-kad-dht/blob/master/src/rpc/handlers/add-provider.js#L17), we assume the sending peer is the provider, ignoring the payload of the message. Should we conform with how [go-libp2p-kad-dht handles this message](https://github.com/libp2p/go-libp2p-kad-dht/blob/master/handlers.go#L362), and ignore the ADD_PROVIDER message if the peer in msg.providerPeers does not match the remote peer id?

The downside is that any client using an older version of js-libp2p-kad-dht will not be able to provide records when connecting to a host running a version of js-libp2p-kad-dht with this fix.

jacobheun:
Logically the code should be:
```
foreach provider in providers
  if provider == peer && provider.multiaddrs > 0
    addProvider(peer)
```

We **only** want to add the peer as a provider if it has multiaddrs and it's the peer sending the request (which should be the only case we get). Right now we're adding the peer as a provider without checking it's multiaddrs which isn't good, because peers shouldn't be providing if they can't be dialed. Ideally we'd also avoid adding them if they just had a basic circuit address (`/p2p-circuit`), but that is a whole other discussion.

We shouldn't ignore the payload because the payload may have multiaddrs we aren't aware of. In theory we should know this from Identify, but that's not a guarantee, and there is nothing prohibiting a peer from filtering its multiaddrs for the dht provide call.

dirkmc:
Sounds good, I can make a PR to implement these changes.

@jacobheun do you see it as a blocker that hosts running an older version of the DHT will not be able to connect to hosts running a version with this fix?

jacobheun:
We could mitigate that by falling back to the from peer if it has multiaddrs and then remove that in a subsequent minor release, or when we do the ipfs dht release.

dirkmc:
Ok will do 👍

PR #129, opened 2019-05-30
Fix: favour providers peerInfo over sender peerInfo in ADD_PROVIDER

dirkmc:
Fixes https://github.com/libp2p/js-libp2p-kad-dht/issues/128


vasco-santos:
@jacobheun I will create that issue and discuss there! But shortly, my idea was to remove those TODO's (from deprecations at least) when we move to async iterators. That way, people using older APIs will only need to change once.

jacobheun:
Ah yeah, that would be a better time to switch, good call.

PR #132, opened 2019-06-03
Async Refactor: query functions are now async functions

kumavis:
This refactor changes the `queryFn`s to be async. The `makePath` functions are still synchronous

```js
const q = new Query(dht, peer.id.id, () => {
  // makePath: do once-per-path setup
  return async () => {
    // queryFn: do actual query
  }
})
```

No behavioral changes were intended, but some tests needed timeouts added to their query functions in order to resolve as expected

related #82

### DIFF
--- a/src/private.js
+++ b/src/private.js
@@ -6,8 +6,6 @@ const timeout = require('async/timeout')
 const PeerInfo = require('peer-info')
 const promisify = require('promisify-es6')
 const promiseToCallback = require('promise-to-callback')
-const callbackify = require('callbackify')
-
 const errcode = require('err-code')
 
 const utils = require('./utils')
@@ -163,7 +161,6 @@ module.exports = (dht) => ({
       record,
       cb
     ))()
-    return undefined
   },
 
   /**
@@ -336,8 +333,7 @@ module.exports = (dht) => ({
       // correct ourself
       if (dht._isSelf(v.from)) {
         try {
-          await promisify(cb => dht._putLocal(key, fixupRec, cb))()
-          // await dht._putLocalAsync(key, fixupRec)
+          await dht._putLocalAsync(key, fixupRec)
         } catch (err) {
           dht._log.error('Failed error correcting self', err)
         }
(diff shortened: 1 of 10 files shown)


test/query.spec.js
```diff
@@ -57,19 +57,19 @@ describe('Query', () => {
     dht.switch.dial = (peer, callback) => callback()
 
     let i = 0
-    const query = (p, cb) => {
+    const query = async (p) => {
```
vasco-santos: why `async` here?

test/query.spec.js
```diff
@@ -90,16 +90,16 @@ describe('Query', () => {
 
     let i = 0
     const visited = []
-    const query = (p, cb) => {
+    const query = async (p) => {
```
vasco-santos: same as before, and other occurrences below that I will not mark in this review

src/private.js
```diff
@@ -463,7 +457,7 @@ module.exports = (dht) => ({
   },
 
   async _verifyRecordOnlineAsync (record) {
-    return promisify(cb => libp2pRecord.validator.verifyRecord(dht.validators, record, cb))()
+    await promisify(cb => libp2pRecord.validator.verifyRecord(dht.validators, record, cb))()
```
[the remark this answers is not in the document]
kumavis: I don't know if there are any significant perf impacts, but i find this more readable

when i see `async` i can clearly tell it returns a promise
when i see `await` instead of a `return` i can clearly see it doesn't return any value

test/query.spec.js
kumavis: this is the `queryFunc` which is an async fn. even if `await syncFn()` works fine, something like `consumePromise(syncFn())` may explode if it does not get a promise

test/query.spec.js
kumavis: see response above

src/private.js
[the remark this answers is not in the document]
vmx: In IPLD I use `async` to signal it returns a promise, but use in those cases `return` (this is also what our currently force you into).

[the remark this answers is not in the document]
vasco-santos: Got it! I typically prefer to have docs for improving readability in those scenarios, but we can iterate on this!

PR #133, opened 2019-06-03
Async Refactor: providers

### LINKED ISSUE #124: dht._findProvidersSingle returns duplicate providers
I frequently see the reponse `msg.providerPeers` contain the same peer many times. I just saw one that had 227 providers where only 5 were unique by `peerId`. weird! :alien: 

I've never seen any duplicates in `msg.closerPeers`

kumavis:
providers.js hasn't been touched in a while, so this is mostly directly taken from #82

Added some async/callback wrappers here and there and used `pull-stream-to-async-iterator` around  `datastore.query` as #82 originally relied on the new datastore interface

Fixes #124

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -537,7 +537,7 @@ class KadDHT extends EventEmitter {
 
     const errors = []
     waterfall([
-      (cb) => this.providers.addProvider(key, this.peerInfo.id, cb),
+      (cb) => promiseToCallback(this.providers.addProvider(key, this.peerInfo.id))(err => cb(err)),
       (cb) => this.getClosestPeers(key.buffer, cb),
       (peers, cb) => {
         const msg = new Message(Message.TYPES.ADD_PROVIDER, key.buffer, 0)
--- a/src/index.js
+++ b/src/index.js
@@ -537,6 +537,7 @@ class KadDHT extends EventEmitter {
 
     const errors = []
     waterfall([
+      // TODO: refactor this in method in async and remove this wrapper
       (cb) => promiseToCallback(this.providers.addProvider(key, this.peerInfo.id))(err => cb(err)),
       (cb) => this.getClosestPeers(key.buffer, cb),
       (peers, cb) => {
--- a/src/providers.js
+++ b/src/providers.js
@@ -2,11 +2,11 @@
 
 const cache = require('hashlru')
 const varint = require('varint')
-const each = require('async/each')
-const pull = require('pull-stream')
-const CID = require('cids')
 const PeerId = require('peer-id')
 const Key = require('interface-datastore').Key
+const Queue = require('p-queue')
+const promisify = require('promisify-es6')
+const toIterator = require('pull-stream-to-async-iterator')
 
 const c = require('./constants')
 const utils = require('./utils')
@@ -56,6 +56,8 @@ class Providers {
     this.lruCacheSize = cacheSize || c.PROVIDERS_LRU_CACHE_SIZE
 
     this.providers = cache(this.lruCacheSize)
+
+    this.syncQueue = new Queue({ concurrency: 1 })
   }
 
   /**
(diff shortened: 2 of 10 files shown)


src/providers.js
```diff
@@ -219,67 +176,55 @@ class Providers {
...
-  addProvider (cid, provider, callback) {
-    this._log('addProvider %s', cid.toBaseEncodedString())
-    const dsKey = makeProviderKey(cid)
-    const provs = this.providers.get(dsKey)
-
-    const next = (err, provs) => {
-      if (err) {
-        return callback(err)
-      }
+  async addProvider (cid, provider) {
```
dirkmc:
```suggestion
  addProvider (cid, provider) {
```

src/providers.js
```diff
@@ -219,67 +176,55 @@ class Providers {
...
    */
-  getProviders (cid, callback) {
-    this._log('getProviders %s', cid.toBaseEncodedString())
-    this._getProvidersMap(cid, (err, provs) => {
-      if (err) {
-        return callback(err)
-      }
-
-      callback(null, Array.from(provs.keys()))
+  async getProviders (cid) {
```
dirkmc:
```suggestion
  getProviders (cid) {
```

src/providers.js
```diff
@@ -219,67 +176,55 @@ class Providers {
...
+      const provs = await this._getProvidersMap(cid)
+      return [...provs.keys()].map((base32PeerId) => {
+        return new PeerId(utils.decodeBase32(base32PeerId))
+      })
     })
   }
 }
 
 /**
  * Encode the given key its matching datastore key.
```
dirkmc:
```suggestion
 * Encode the given CID as a datastore provider key.
```

src/providers.js
```diff
@@ -289,48 +234,59 @@ function makeProviderKey (cid) {
...
  * @param {PeerId} peer
  * @param {number} time
- * @param {function(Error)} callback
- * @returns {undefined}
+ * @returns {Promise}
  *
  * @private
  */
-function writeProviderEntry (store, cid, peer, time, callback) {
+async function writeProviderEntry (store, cid, peer, time) {
```
dirkmc:
```suggestion
function writeProviderEntry (store, cid, peer, time) {
```

test/rpc/handlers/get-providers.spec.js
```diff
@@ -89,7 +90,7 @@ describe('rpc - handlers - GetProviders', () => {
 
     waterfall([
       (cb) => dht._add(closer, cb),
-      (cb) => dht.providers.addProvider(v.cid, prov, cb),
+      (cb) => promiseToCallback(dht.providers.addProvider(v.cid, prov))(err => cb(err)),
```
dirkmc:
```suggestion
      (cb) => promiseToCallback(dht.providers.addProvider(v.cid, prov))(cb),
```

src/index.js
```diff
@@ -537,7 +537,7 @@ class KadDHT extends EventEmitter {
 
     const errors = []
     waterfall([
-      (cb) => this.providers.addProvider(key, this.peerInfo.id, cb),
+      (cb) => promiseToCallback(this.providers.addProvider(key, this.peerInfo.id))(err => cb(err)),
```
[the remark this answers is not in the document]
kumavis: this is intentional as it can screw up `async/waterfall` by passing in an `undefined` result.

src/providers.js
kumavis: here's what I wrote in #134 :
> I don't know if there are any significant perf impacts, but i find this more readable

> when i see async i can clearly tell it returns a promise
> when i see await instead of a return i can clearly see it doesn't return any value

kumavis:
@dirkmc committed the comment changes
The other points of feedback refer intentional to decisions I made, and I responded to the first instance of each

src/index.js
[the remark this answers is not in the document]
kumavis: I agree its ugly, and look forward to unwrapping it in my next async refactor pass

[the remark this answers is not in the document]
vasco-santos: Could you just please add a comment explaining that with a `TODO` for being refactored?

test/providers.spec.js
```diff
@@ -4,165 +4,147 @@
...
+      providers.addProvider(cid, infos[0].id),
+      providers.addProvider(cid, infos[1].id)
+    ])
+
+    const provs = await providers.getProviders(cid)
+    const ids = new Set(provs.map((peerId) => peerId.toB58String()))
+    expect(ids.has(infos[0].id.toB58String())).to.be.eql(true)
+    expect(ids.has(infos[1].id.toB58String())).to.be.eql(true)
+
+    providers.stop()
```
vasco-santos: Add and after with:

```js
providers && providers.stop()
```

And you can remove this from all the tests

vasco-santos:
Also, could you increase the bundlesize max size in `.aegir`, in order to have a green CI?

kumavis:
@vasco-santos @jacobheun ready to go

PR #134, opened 2019-06-03
Async Refactor: random-walk

kumavis:
switched over to a `_running` state var and and `while(true)` async loop instead of managing `timeoutId`s

skipped a test that was expecting queries to be run in series and that errors would stop the other queries. Should an error in `_walk` trigger `randomWalk` to stop? currently we log and ignore

updates #82

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,6 +54,8 @@
     "merge-options": "^1.0.1",
     "multihashes": "~0.4.14",
     "multihashing-async": "~0.5.2",
+    "p-queue": "^5.0.0",
+    "p-times": "^2.1.0",
     "peer-id": "~0.12.2",
     "peer-info": "~0.15.1",
     "priorityqueue": "~0.2.1",
--- a/src/random-walk.js
+++ b/src/random-walk.js
@@ -69,9 +69,8 @@ class RandomWalk {
    * @private
    */
   async _runPeriodically () {
-    while (true) {
-      // exit if the walk has been stopped
-      if (!this._timeoutId) return
+      // run until the walk has been stopped
+      while (this._timeoutId) {
       try {
         await this._walk(this._options.queriesPerPeriod, this._options.timeout)
       } catch (err) {
--- a/src/random-walk.js
+++ b/src/random-walk.js
@@ -28,7 +28,7 @@ class RandomWalk {
     this._options = { ...c.defaultRandomWalk, ...options }
     this._kadDHT = dht
     this.log = logger(dht.peerInfo.id, 'random-walk')
-    this._running = false
+    this._timeoutId = undefined
   }
 
   /**
@@ -40,12 +40,10 @@ class RandomWalk {
    */
   start () {
     // Don't run twice
-    if (this._running || !this._options.enabled) { return }
-
-    this._running = true
+    if (this._timeoutId || !this._options.enabled) { return }
 
     // Start doing random walks after `this._options.delay`
-    setTimeout(() => {
+    this._timeoutId = setTimeout(() => {
       // Start runner immediately
       this._runPeriodically()
     }, this._options.delay)
@@ -58,7 +56,10 @@ class RandomWalk {
    * @returns {void}
    */
   stop () {
-    this._running = false
+    if (this._timeoutId) {
+      clearTimeout(this._timeoutId)
+      this._timeoutId = undefined
+    }
     this._controller && this._controller.abort()
   }
 
@@ -69,14 +70,17 @@ class RandomWalk {
    */
   async _runPeriodically () {
     while (true) {
-      if (!this._running) return
+      // exit if the walk has been stopped
+      if (!this._timeoutId) return
       try {
         await this._walk(this._options.queriesPerPeriod, this._options.timeout)
       } catch (err) {
         this._kadDHT._log.error('random-walk:error', err)
       }
       // Each subsequent walk should run on a `this._options.interval` interval
-      await new Promise(resolve => setTimeout(resolve, this._options.interval))
+      await new Promise(resolve => {
+        this._timeoutId = setTimeout(resolve, this._options.interval)
+      })
     }
   }
 
--- a/src/random-walk.js
+++ b/src/random-walk.js
@@ -69,8 +69,8 @@ class RandomWalk {
    * @private
    */
   async _runPeriodically () {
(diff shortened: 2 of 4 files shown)


src/random-walk.js
```diff
@@ -59,70 +58,67 @@ class RandomWalk {
...
    */
-  _runPeriodically (walk, interval) {
-    this._timeoutId = setTimeout(() => {
-      walk((nextInterval) => {
-        // Schedule next
-        this._runPeriodically(walk, nextInterval)
-      })
-    }, interval)
+  async _runPeriodically () {
+    while (true) {
```
dirkmc:
```suggestion
    while (this._running) {
```

src/random-walk.js
```diff
@@ -59,70 +58,67 @@ class RandomWalk {
...
+      if (!this._running) return
+      try {
+        await this._walk(this._options.queriesPerPeriod, this._options.timeout)
+      } catch (err) {
+        this._kadDHT._log.error('random-walk:error', err)
+      }
+      // Each subsequent walk should run on a `this._options.interval` interval
+      await new Promise(resolve => setTimeout(resolve, this._options.interval))
```
dirkmc: Can we use `Promise.race()` here to detect when the random walk has been stopped? Otherwise we may need to wait a full interval when shutting down the DHT. (Note that you may need to use an event that gets fired on stop)

dirkmc:
I don't think an error should cause random walk to stop - we don't really care about past errors for the purposes of random walk, we just want to walk the DHT to find new peers

src/random-walk.js
kumavis: @dirkmc are you just concerned about the timeout keeping a node process alive?
if so we can [unref it](https://nodejs.org/api/timers.html#timers_timeout_unref)
otherwise it shouldnt be an issue

I guess one problem could be if `stop` and `start` are called during the interval, we'll end up with two loops running...

kumavis:
@dirkmc i `skip()`ed a test that expects an error to stop the query, do you think i should delete the test?

src/random-walk.js
dirkmc: That's true I guess you can just keep a reference and call clearTimeout()

dirkmc:
I would say we should have a test for the opposite case - that the random walk *does* continue even if there's an error, but I would like to hear from @vasco-santos and @jacobheun if they know of a reason why it was written this way originally.

jacobheun:
> I would say we should have a test for the opposite case - that the random walk does continue even if there's an error.

I agree, there's no reason to stop the random walk on an error. The only reason the walk should stop is if the query is done or it timed out.

vasco-santos:
Agree that it should keep going!

kumavis:
@vasco-santos @jacobheun ready to go

kumavis:
fixed the conflict

package.json
```diff
@@ -55,6 +55,7 @@
     "multihashes": "~0.4.14",
     "multihashing-async": "~0.5.2",
```
vasco-santos: Can you update [multiformats/js-multihashing-async](https://github.com/multiformats/js-multihashing-async) to the latest version, as it will already support promises? It is also used in utils, but you can callbackify those utils for now (address utils in an upcoming PR)

src/random-walk.js
```diff
@@ -1,16 +1,15 @@
 'use strict'
 
-const times = require('async/times')
+const promisify = require('promisify-es6')
 const crypto = require('libp2p-crypto')
-const waterfall = require('async/waterfall')
-const multihashing = require('multihashing-async')
+const multihashing = promisify(require('multihashing-async'))
```
vasco-santos: If updated, `promisify` will not be needed here

src/random-walk.js
```diff
@@ -139,44 +138,41 @@ class RandomWalk {
...
-      // wait what, there was something found? Lucky day!
-      callback(errcode(new Error(`random-walk: ACTUALLY FOUND PEER: ${peer}, ${id.toB58String()}`), 'ERR_FOUND_RANDOM_PEER'))
-    })
+      throw err
+    }
+
+    this.log('query:found', peer)
+
+    // wait what, there was something found? Lucky day!
+    throw errcode(new Error(`random-walk: ACTUALLY FOUND PEER: ${peer}, ${id.toB58String()}`), 'ERR_FOUND_RANDOM_PEER')
```
vasco-santos:
```suggestion
    throw errcode(`random-walk: ACTUALLY FOUND PEER: ${peer}, ${id.toB58String()}`, 'ERR_FOUND_RANDOM_PEER')
```

package.json
kumavis: @vasco-santos happy to do that in a separate PR

kumavis:
@vasco-santos I'll handle the `multihashing-async` update in a separate PR, as it will affect more files than `randomWalk`

src/random-walk.js
```diff
@@ -59,70 +56,72 @@ class RandomWalk {
...
-        this._runPeriodically(walk, nextInterval)
+  async _runPeriodically () {
+    // run until the walk has been stopped
+    while (this._timeoutId) {
+      try {
+        await this._walk(this._options.queriesPerPeriod, this._options.timeout)
+      } catch (err) {
+        this._kadDHT._log.error('random-walk:error', err)
+      }
+      // Each subsequent walk should run on a `this._options.interval` interval
```
dirkmc: Should we be checking `this._timeoutId` has not been cleared after the walk has completed?

src/random-walk.js
```diff
@@ -59,70 +56,72 @@ class RandomWalk {
...
    */
-  _walk (queries, walkTimeout, callback) {
+  async _walk (queries, walkTimeout) {
     this.log('start')
     this._controller = new AbortController()
 
-    times(queries, (i, next) => {
-      this.log('running query %d', i)
+    try {
+      await times(queries, async (index) => {
```
dirkmc: Should we be checking if the DHT has been stopped each time we run the query?

vasco-santos: As we only use `_walk` from  `_runPeriodically`, which immediately verifies if the dht has stopped and we run the queries concurrently, do you feel that we need this here?

dirkmc: Ah I didn't realize they run concurrently 👍

PR #135, opened 2019-06-25
Docs: implementation summary

dirkmc:


docs/IMPL_SUMMARY.MD
```diff
@@ -0,0 +1,65 @@
+# js-libp2p-kad-dht
+
+js-libp2p-kad-dht is a JavaScript implementation of the Kademlia DHT with some features of S/Kademlia. A "provider" node uses the DHT to advertise that it has a particular piece of content, and "querying" nodes can search the DHT for peers that have a particular piece of content. Content is modeled as a value that is identified by a key, where the key and value are Buffers.
```
vasco-santos:
```suggestion
js-libp2p-kad-dht is a JavaScript implementation of the Kademlia DHT with some features of S/Kademlia. A "provider" node uses the DHT to advertise that it has a particular piece of content, and "querying" nodes will search the DHT for peers that have a particular piece of content. Content is modeled as a value that is identified by a key, where the key and value are Buffers.
```

docs/IMPL_SUMMARY.MD
```diff
@@ -0,0 +1,65 @@
...
+#### GET value
+
+`get (Key): [Value]`
+
+To retrieve a value from the DHT
+- the `querying node` converts the key to a DHT id
+- the `querying node` follows the "closest peers" algorithm to find the nearest peers to the DHT id
+- at each iteration of the algorithm, if the peer has the value it responds with the value itself in addition to closer peers.
+
+Note that because the DHT is asynchronous, the value for a particular key may be different at each node, so `GET value` may return distinct values. The DHT has "selectors", configurable functions that choose the "best" value (eg the most recently modified).
```
vasco-santos:
```suggestion
Note that because the DHT is asynchronous, the value for a particular key may be different at each node (not synchronized), so `GET value` may return distinct values. The DHT has "selectors", configurable functions that choose the "best" value (eg the most recently modified).
```

dirkmc:
@olizilla could you please take a look at this doc and let me know where you think things could be clearer from the perspective of someone who is not familiar with the internal workings of the internals of the DHT

@michaelavila could you please take a look also

Thanks ❤️

docs/IMPL_SUMMARY.MD
```diff
@@ -0,0 +1,65 @@
...
+#### GET value
+
+`get (Key): [Value]`
+
+To retrieve a value from the DHT
+- the `querying node` converts the key to a DHT id
+- the `querying node` follows the "closest peers" algorithm to find the nearest peers to the DHT id
+- at each iteration of the algorithm, if the peer has the value it responds with the value itself in addition to closer peers.
+
+Note that because the DHT is asynchronous, the value for a particular key may be different at each node (not synchronized), so `GET value` may return distinct values. The DHT has "selectors", configurable functions that choose the "best" value (eg the most recently modified).
```
michaelavila: This paragraph isn't entirely clear to me. Is this saying that `GET key` may return two different `value`, depending on which node services the request? Is this talking about the routing records or the actual value produced when attempting to get a key?

dirkmc: Right, basically it's saying that each node has its own copy of the value, and these may get updated asynchronously, so they may not all be in sync. This is talking about the actual value stored by the PUT RPC. For example in the case of IPNS this would be be the IPNS record itself in serialized form.

Is this clearer?

> Note that the value for a particular key is stored by many nodes, and these nodes receive `PUT` requests asynchronously, so it's possible that nodes may have distinct values for the same key. For example if node A `PUT`s the value `hello` to key `greeting` and node B concurrently `PUT`s the value `bonjour` to key `greeting`, some nodes close to the key `greeting` may receive `hello` first and others may receive `bonjour` first.

> Therefore a `GET` request to the DHT may collect distinct values (eg `hello` and `bonjour`) for a particular key from the nodes close to the key. The DHT has "selectors", configurable functions that choose the "best" value (for example IPNS records include a sequence number, and the "best" value is the record with the highest sequence number).

vasco-santos:
This looks good to me!

@michaelavila @olizilla all ok for you to get this merged?

vasco-santos:
I will merge this PR, as the current information is better than having nothing. If anyone thinks we should add or modify anything, we can do that in a new PR!

PR #137, opened 2019-07-24
fix _findNProvidersAsync discarding search results

Qmstream:
When `_findNProvidersAsync` times out, it ignores the search results, as it's checking the `out` array before it's actually populated.

### DIFF
--- a/src/private.js
+++ b/src/private.js
@@ -563,7 +563,7 @@ module.exports = (dht) => ({
         promiseToCallback(query.run(peers))(cb)
       }, providerTimeout)(callback))()
     } catch (err) {
-      if (err.code !== 'ETIMEDOUT' || out.length === 0) {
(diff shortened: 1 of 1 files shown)


Qmstream:
@jacobheun, I removed the dangling white space.

But, sorry, I have no idea how I could write the suggested test (I have little experience with this).

jacobheun:
@Qmstream no problem, I will look at getting a test added for this. I added https://github.com/libp2p/js-libp2p-kad-dht/issues/138 to track getting the test added so we don't block this, since the test addition isn't critical and it would be nice to have this out sooner.

vasco-santos:
Thanks for looking to the PR and creating the issue @jacobheun

vasco-santos:
Also @Qmstream thanks for this PR ❤️

PR #139, opened 2019-07-31
filter out invalid records

Qmstream:
(As requested: https://github.com/ipfs/js-ipfs/issues/2183#issuecomment-515891676.)

`getMany` is returning a list of results with empty values for invalid records, but `_getAsync` and `_sendCorrectionRecord` weren't checking for them.

### DIFF
--- a/src/private.js
+++ b/src/private.js
@@ -287,7 +287,7 @@ module.exports = (dht) => ({
 
     const vals = await promisify(cb => dht.getMany(key, c.GET_MANY_RECORD_COUNT, options, cb))()
 
-    const recs = vals.map((v) => v.val)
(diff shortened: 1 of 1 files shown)


vasco-santos:
Hey @Qmstream thanks for this PR. 

I was thinking why `getMany` is sending empty values for invalid records, as I would see that being fixed a better approach. Did you get any conclusion on that? Otherwise, I will have to take a deeper look

Qmstream:
From what I understand, it's in order to track the peers with invalid values, so as to send them the correction value later.

`_getAsync` calls `getMany` to gather up to `GET_MANY_RECORD_COUNT` peers with (good & bad) values, selects the best value, and sends it back to those peers.

Issue #111, opened 2019-05-08
randomWalk start/stop does not take callbacks


kumavis:
`randomWalk.start()` and `randomWalk.stop()` should take callbacks
https://github.com/libp2p/js-libp2p-kad-dht/blob/master/src/random-walk.js#L41

https://github.com/libp2p/interface-peer-discovery#api

Edit: interface has been updated to use promises instead of callbacks

kumavis:
`libp2p-webrtc-star` **does** implement this correctly
https://github.com/libp2p/js-libp2p-webrtc-star/blob/master/src/index.js#L46-L47

vasco-santos:
Closing this in favour of #44 that should be addressed once we refactor the DHT with the new spec.

PR #141, opened 2019-08-24
fix: match libp2p/interface-peer-discovery spec

### LINKED ISSUE #44: randomWalk should be tested through interface-peer-discovery tests
The current `randomWalk` implementation is not being tested with the [libp2p/interface-peer-discovery](https://github.com/libp2p/interface-peer-discovery).

This should have a similar implementation suite of tests like the other `peer-discovery` modules.

andycodesstuff:
The following pull requests fixes #44 and #111 by matching the `libp2p/interface-peer-discovery` spec.

### DIFF
--- a/src/random-walk.js
+++ b/src/random-walk.js
@@ -8,6 +8,7 @@ const assert = require('assert')
 const AbortController = require('abort-controller')
 const errcode = require('err-code')
 const times = require('p-times')
+const setImmediate = require('async/setImmediate')
 const c = require('./constants')
 const { logger } = require('./utils')
 
@@ -36,9 +37,10 @@ class RandomWalk {
    * every interval requesting random data. This is done to keep the dht
    * healthy over time.
    *
+   * @param {function()} callback
    * @returns {void}
    */
-  start () {
+  start (callback) {
     // Don't run twice
     if (this._timeoutId || !this._options.enabled) { return }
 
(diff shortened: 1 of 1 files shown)


andrycodestuffs:
Hope to finish this as soon as possible, I'll keep you updated!

vasco-santos:
Cool, really glad that you are taking this. Bear in mind that we are also working on moving from callback into Promises, which may reflect on this: [ipfs/js-ipfs#1670](https://github.com/ipfs/js-ipfs/issues/1670)

andrycodestuffs:
@vasco-santos could you please provide a very small example of how is random walk discovery implemented at the moment?
I'd like to have a reference point to understand the current execution flow

vasco-santos:
At the moment, random-walk is started and stopped within the `DHT` codebase, if it is enabled through the received options:

https://github.com/libp2p/js-libp2p-kad-dht/blob/master/src/index.js#L170
https://github.com/libp2p/js-libp2p-kad-dht/blob/master/src/index.js#L184

Briefly, the `random-walk` receives the `dht` instance (https://github.com/libp2p/js-libp2p-kad-dht/blob/master/src/index.js#L136) and uses it to create queries in the DHT. 

When a peer is found via those queries, a event is fired https://github.com/libp2p/js-libp2p-kad-dht/blob/30fd44a5780e6ca9bad3387478c0643a27ecce3b/src/index.js#L699 with the discovered peer, which should be inside the `random-walk` scope. 

As a result, we should be able to change the configuration to use it separately. Taking an example on the `libp2p` configuration in `js-ipfs` (https://github.com/ipfs/js-ipfs/blob/master/src/core/runtime/libp2p-nodejs.js#L43-L48), we should be able to add a random-walk instance inside `peerDiscovery`. Some stuff needs to change in the configuration side of things too, as we need to create the DHT instance in advance, in order to create the `random-walk`.

Also note that the `random-walk` needs to be exported to be used in the `js-libp2p` scope

achingbrain:
Thanks for taking the time to open this, I'm sorry it never landed.  Random walk has been removed from the implementation as it can make queries now instead.

PR #143, opened 2019-08-26
Allow secp256k1 public keys as PeerIds

robertkiel:
Why?
- using secp256k1 pub keys is more convenient than using hashed pubkeys

What has changed?
- setting 'peer-id' to `~0.12.3` to support secp256k1 pubkey-based peerIds


package.json
```diff
@@ -49,15 +49,15 @@
...
-    "multihashes": "~0.4.14",
-    "multihashing-async": "~0.5.2",
+    "multihashes": "~0.4.15",
+    "multihashing-async": "~0.6.0",
     "p-queue": "^5.0.0",
     "p-times": "^2.1.0",
-    "peer-id": "~0.12.2",
-    "peer-info": "~0.15.1",
+    "peer-id": "~0.12.3",
+    "peer-info": "~0.16.0",
```
jacobheun: Several of the minor version bumps here are from callbacks to async/await, which are breaking changes. I think most of these should be promisified calls, they will need to be updated to no longer use promisify and use await calls instead.

robertkiel: It works also with `peer-info 0.15.1` which comes without promisifications

jacobheun:
This PR is out of date, all deps are up to date in master.

PR #148, opened 2019-11-07
refactor: async await

vasco-santos:
In the context of [ipfs/js-ipfs#1670](https://github.com/ipfs/js-ipfs/issues/1670) all the API is moving into async

BREAKING CHANGE: Switch to using async/await and async iterators.

PR #153, opened 2019-11-20
refactor: async it

vasco-santos:
This PR aims to use `async iterators` instead of `pull-streams` in the context of the asyn migration, as a follow up for #148


package.json
```diff
@@ -57,26 +60,22 @@
...
+    "async-iterator-all": "^1.0.0",
     "chai": "^4.2.0",
     "datastore-level": "~0.12.1",
     "delay": "^4.3.0",
     "dirty-chai": "^2.0.1",
-    "interface-connection": "~0.3.3",
-    "libp2p-mplex": "~0.8.5",
-    "libp2p-switch": "~0.42.7",
-    "libp2p-tcp": "~0.13.0",
+    "it-pair": "^1.0.0",
```
vasco-santos: I will create an issue to track the removal of `peer-book` for tests, as well as handle the `peer-store` operations currently done here.

test/multiple-nodes.spec.js
```diff
@@ -100,10 +96,10 @@ describe('multiple nodes', () => {
     await dhts[4].put(key, Buffer.from('world4'))
 
     const res = await Promise.all([
-      dhts[3].get(key, { maxTimeout: 2000 }),
```
vasco-santos: I dont't know as this was passing before. 

It was getting stored locally `world3` (after put), which was the first result obtained from `getMany`. Then, we use the `selector: () => 0` and the first record was obtained

src/index.js
```diff
@@ -63,14 +67,31 @@ class KadDHT extends EventEmitter {
     randomWalk = {}
   }) {
     super()
-    assert(sw, 'libp2p-kad-dht requires a instance of Switch')
+    assert(dialer, 'libp2p-kad-dht requires a instance of Dialer')
```
dirkmc:
```suggestion
    assert(dialer, 'libp2p-kad-dht requires an instance of Dialer')
```

src/index.js
```diff
@@ -63,14 +67,31 @@ class KadDHT extends EventEmitter {
...
+    this.dialer = dialer
+
+    /**
+     * Local peer info
+     * @type {PeerInfo}
+     */
+    this.peerInfo = peerInfo
+
+    /**
+     * Local peer info
```
dirkmc:
```suggestion
     * Local PeerStore
```

src/network.js
```diff
@@ -170,47 +173,35 @@ class Network {
...
-      })
-    )
-  })
+async function writeReadMessage (conn, msg) {
+  const res = await pipe(
+    [msg],
+    lp.encode(),
+    conn,
+    utils.itFilter(
+      (msg) => msg.length < c.maxMessageSize
```
dirkmc: Does this mean that messages that are larger that `c.maxMessageSize` will be discarded without an error message? Maybe we should open an issue to change that behaviour (in a separate PR)

src/network.js
```diff
@@ -170,47 +173,35 @@ class Network {
...
+    ),
+    lp.decode(),
+    async source => {
+      for await (const chunk of source) {
```
dirkmc: Not sure if it's worth adding a dependency just for this, but the streaming iterables library has some useful functions, eg this could be replaced with a [map](https://github.com/bustle/streaming-iterables#map) and a collect

src/rpc/index.js
```diff
@@ -14,24 +16,23 @@ module.exports = (dht) => {
...
-   *
    * @param {PeerInfo} peer
    * @param {Message} msg
    * @returns {Promise<Message>}
    *
    * @private
    */
-  async function handleMessage (peer, msg) {
+  async function handleMessage (peer, msg) { // eslint-disable-line
+    // get handler & exectue it
```
dirkmc:
```suggestion
    // get handler & execute it
```

src/rpc/index.js
```diff
@@ -41,63 +42,43 @@ module.exports = (dht) => {
...
-        pull.filter(Boolean),
-        lp.encode(),
-        conn
-      )
-    })
+    await pipe(
+      stream.source,
+      lp.decode(),
+      utils.itFilter(
+        (msg) => msg.length < c.maxMessageSize
```
dirkmc: Same as above - we should consider doing more than failing silently

src/utils.js
```diff
@@ -11,6 +11,16 @@ const { Record } = require('libp2p-record')
 const PeerId = require('peer-id')
 const errcode = require('err-code')
 
+exports.itFilter = function (predicate) {
```
dirkmc: Consider using https://github.com/bustle/streaming-iterables#filter

test/query/index.spec.js
```diff
@@ -46,11 +46,14 @@ describe('Query', () => {
...
 
     before('create a dht', () => {
+      const peerStore = new PeerBook()
       dht = new DHT({
-        sw: {
+        dialer: {
           _peerInfo: ourPeerInfo,
-          _peerBook: new PeerBook()
-        }
+          _peerBook: peerStore
```
jacobheun: `dialer._peerBook` and `dialer._peerInfo` should be deleted, they're not properties of the dialer.

test/rpc/handlers/get-value.spec.js
```diff
@@ -87,6 +87,7 @@ describe('rpc - handlers - GetValue', () => {
 
       const msg = new Message(T, key, 0)
 
+      dht.peerStore.put(other)
```
jacobheun: Did this need to get added?

test/rpc/index.spec.js
```diff
@@ -4,69 +4,82 @@
...
+    const data = []
+    await pipe(
+      [msg.serialize()],
+      lp.encode(),
+      async source => {
+        for await (const chunk of source) {
+          data.push(chunk.slice())
+        }
+      }
+    )
```
jacobheun: If you use `collect` from `streaming-iterables` I think this code can be simplified a bit (as well as some of the other places we are doing collects here. Also, there are multiple slices of the BufferList happening in this stream, it's not needed. You could also use the `it-buffer` module and simplify all of this to be:

```js
    const data = await pipe(
      [msg.serialize()],
      lp.encode(),
      collect
    )

    const duplexStream = {
      source: data,
      sink: async (source) => {
        const res = await pipe(
          source,
          lp.decode(),
          toBuffer,  // Ensure we have buffers here for validateMessage to consume
          collect
        )
        validateMessage(res)
      }
    }
```

test/utils/index.js
```diff
@@ -3,17 +3,54 @@
 const delay = require('delay')
 const pRetry = require('p-retry')
 const pTimeout = require('p-timeout')
-const promisify = require('promisify-es6')
+const DuplexPair = require('it-pair/duplex')
```
jacobheun: just for consistency
```suggestion
const duplexPair = require('it-pair/duplex')
```

test/utils/index.js
```diff
@@ -3,17 +3,54 @@
...
+  },
+  unregister: (id) => {
+    delete registrarRecord[id]
+  }
+})
+
+exports.createMockRegistrar = createMockRegistrar
+
+const ConnectionPair = () => {
+  const [d0, d1] = DuplexPair()
```
jacobheun:
```suggestion
  const [d0, d1] = duplexPair()
```

package.json
vasco-santos: we still use `async/queue` for the query `WorkerQueue`. I will create an issue to track that

src/network.js
vasco-santos: sounds good, can you create it?

dirkmc: https://github.com/libp2p/js-libp2p-kad-dht/issues/154

test/rpc/index.spec.js
vasco-santos: yap, cool tip!

test/rpc/handlers/get-value.spec.js
vasco-santos: Yes, in the code we try to find the peer in the `peer-store`. After this test we have another one for unkwown peers

PR #158, opened 2019-12-02
fix: peerstore get and has param

vasco-santos:
The new `peer-store` for `get` and `has` should receive the `b58str` and not `peer-id` as https://github.com/libp2p/js-libp2p/blob/refactor/async-await/src/peer-store/index.js#L163

This module is still using `js-peerbook` for tests (for mock peerstore) which accepts both `b58str` and `peer-id` [libp2p/js-peer-book#getpeeridlike](https://github.com/libp2p/js-peer-book#getpeeridlike), which made the tests here pass. 

Anyway, I will create an issue to create a proper mock for the tests here, so that we can deprecate `peer-book`

### DIFF
--- a/src/content-routing/index.js
+++ b/src/content-routing/index.js
@@ -81,8 +81,8 @@ module.exports = (dht) => {
 
       provs.forEach((id) => {
         let info
-        if (dht.peerStore.has(id)) {
-          info = dht.peerStore.get(id)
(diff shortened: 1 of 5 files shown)


vasco-santos:
After discussing with @jacobheun we will support only `peer-id` as a param, as we plan to move to `base32` id's by default in the future. I will create a PR for supporting `peer-id` on the `peerstore` instead.

Issue #164, opened 2019-12-12
for-await handled serially


kumavis:
https://github.com/libp2p/js-libp2p-kad-dht/commit/f0e68004b8c559c3214a1e0aaa94a15a26826085#diff-463b2787d6b1decd9b30c6a23ef7aa95

@vasco-santos 
If I'm understanding this correctly, with this new AsyncIterator and for-await syntax we:
- we start messaging our first dht peer sooner (yay!)
- message subsequent peers serially after awaiting a response (boo!)

Seems like we want to do the message sending in parallel, so we'd want some async parallel queue ( maybe https://github.com/dbushong/async-q#queue )

Does that make sense? Need a good general solution here, seems common to want to chain async actions originating from an AsyncIterator and handle them all in parallel

vasco-santos:
@kumavis the first step here was to have an async iterable being returned from now. 

Internally, we are still not leveraging the real capabilities of `async iterators` for the majority of the cases yet. So you are right that the end goal is to handle that stuff in parallel as we get them within the `for await` loop. Now that we expose this API externally and release the async refactor we intend to come back to the DHT soon.

One of the things that I have in mind (short term) is that, as well as to be able to use `.get` in an async iterable fashion (might replace getMany, and the user can iterate while he wants).

Hopefully, I will be able to dedicate more time to the `dht` next year, as we also want to rethink some stuff and improve the DHT in the browser runtimes.

kumavis:
Sounds good!
Looks like `paramap-it` solves exactly this problem.

https://github.com/alanshaw/paramap-it

kumavis:
Ah, while `paramap-it` should help we shouldn't need it's order preservation.

vasco-santos:
yeah, you can have a look at [it-awesome](https://github.com/alanshaw/it-awesome). It has a lot of great modules we can use with async iterators

PR #166, opened 2019-12-12
fix: use utils.mapParallel for parallel processing of peers

kumavis:
Fixes https://github.com/libp2p/js-libp2p-kad-dht/issues/164

There's still one more place where we handle things serially: handling incoming messages for a single peer. I decided not to touch it. It could be addressed with [`paramap-it`](https://github.com/alanshaw/paramap-it), likely with `ordered: false` option

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -206,3 +206,22 @@ exports.withTimeout = (asyncFn, time) => {
     ])
   }
 }
+
+/**
+ * Iterates the given `asyncIterator` and runs each item through the given `asyncFn` in parallel.
+ * Returns a promise that resolves when all items of the `asyncIterator` have been passed
+ * through `asyncFn`.
+ *
+ * @param {AsyncIterable} [asyncIterator]
+ * @param {Function} [asyncFn]
+ * @returns {Array}
+ *
+ * @private
+ */
+exports.mapParallel = async function (asyncIterator, asyncFn) {
+  const tasks = []
+  for await (const item of asyncIterator) {
+    tasks.push(asyncFn(item))
+  }
+  return Promise.all(tasks)
+}
--- a/src/content-fetching/index.js
+++ b/src/content-fetching/index.js
@@ -109,7 +109,7 @@ module.exports = (dht) => {
       let counterAll = 0
       let counterSuccess = 0
 
-      for await (const peer of dht.getClosestPeers(key, { shallow: true })) {
+      await utils.mapParallel(dht.getClosestPeers(key, { shallow: true }), async (peer) => {
         try {
           counterAll += 1
           await dht._putValueToPeer(key, record, peer)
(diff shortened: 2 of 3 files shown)


src/utils.js
```diff
@@ -205,3 +205,22 @@ exports.withTimeout = (asyncFn, time) => {
...
+ * @returns {Array}
+ *
+ * @private
+ */
+exports.mapParallel = async function (asyncIterator, asyncFn) {
+  const tasks = []
+  for await (const item of asyncIterator) {
+    tasks.push(asyncFn(item))
+  }
+  return Promise.all(tasks)
```
alanshaw: We have to wait for `asyncIterator` to end before parallel mapping over the items it yields. This is a little unusual and potentially buffers a large amount of data in memory.

alanshaw:
Worth mentioning and maybe using here because it has concurrency limits:
https://www.npmjs.com/package/streaming-iterables#parallelmap

src/utils.js
kumavis: > We have to wait for asyncIterator to end before parallel mapping over the items it yields

i dont understand what you mean. that is neither a requirement we have nor the behavior of this code.

each item starts its `asyncFn` immediately when yielded. handling of each `asyncFn` does not block further iteration. In both my uses all I needed was to know when all have completed, so I returned a `Promise.all`

kumavis:
> Is there some performance bottleneck that this is addressing? 

no, this was trying to fix an architectural regression I noticed where network activity became done in serial while it was previously done in parallel. (though previously it waited for all peers before processing).

adding a concurrency limit seems reasonable

I expect performance to be blocked primarily by network i/o, so eagerly getting messages out seems like the correct answer, and doing it in series seems like a very questionable choice. If its intentionally time inefficient, it should have a comment explaining that.

kumavis:
Here's a quick experiment testing mapParallel perf vs the previous behavior. if iterating and the task take the same amount of time, its 2x fast. Significantly faster if asyncFn is slower. Marginally faster if iterator is slower.
https://gist.github.com/kumavis/3032e17c351e4e1f4874922d0481de09

I guess one concern is if asyncFn rejects before we call `Promise.all`, it would be considered and unhandled promise

kumavis:
I guess one concern is if asyncFn rejects before we call Promise.all, it would be considered and unhandled promise

src/utils.js
alanshaw: ah you’re correct - my apologies I didn’t notice the function call!

vasco-santos:
> I guess one concern is if asyncFn rejects before we call Promise.all, it would be considered and unhandled promise

what if we use `p-queue`? we can also control concurrency

vasco-santos:
Hey @kumavis . Should we get this to the finish line? Or include it when we work on updating this to conform with the new updates in the go implementation of the dht?

kumavis:
work is done, hit that merge. or add your own commits if you want the p-queue

vasco-santos:
FYI I just rebased this PR and waiting on CI

vasco-santos:
FYI I will consider adding the `p-queue` when we refactor the dht

PR #179, opened 2020-04-01
chore: use new peer store api

vasco-santos:
In the context of the PeerStore improvements specified on [libp2p/js-libp2p#582](https://github.com/libp2p/js-libp2p/issues/582) and as a follow up of [libp2p/js-libp2p#590](https://github.com/libp2p/js-libp2p/pull/590), this PR switches `libp2p-kad-dht` usage of `peer-store` to its new API.

Moreover, there were some places where data was being added to the `peerStore` with any benefit. In some cases the data was already there and in other cases there was nothing new to add at that point.

BREAKING CHANGE: uses new peer-store api

Needs:

- [x] [libp2p/js-libp2p#590](https://github.com/libp2p/js-libp2p/pull/590)

Unblocks:

- [ ] [libp2p/js-libp2p#598](https://github.com/libp2p/js-libp2p/pull/598)

### DIFF
--- a/package.json
+++ b/package.json
@@ -76,6 +76,7 @@
     "delay": "^4.3.0",
     "dirty-chai": "^2.0.1",
     "it-pair": "^1.0.0",
+    "libp2p": "libp2p/js-libp2p#chore/deprecate-old-peer-store-api",
     "lodash": "^4.17.11",
     "lodash.random": "^3.2.0",
     "lodash.range": "^3.2.0",
@@ -83,7 +84,6 @@
     "p-each-series": "^2.1.0",
     "p-map-series": "^2.1.0",
     "p-retry": "^4.2.0",
-    "peer-book": "~0.9.2",
     "sinon": "^9.0.0"
   },
   "contributors": [
--- a/src/peer-routing/index.js
+++ b/src/peer-routing/index.js
@@ -4,7 +4,6 @@ const errcode = require('err-code')
 const pTimeout = require('p-timeout')
 
 const PeerId = require('peer-id')
-const PeerInfo = require('peer-info')
 const crypto = require('libp2p-crypto')
 
 const c = require('../constants')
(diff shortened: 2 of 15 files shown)


package.json
```diff
@@ -76,14 +76,14 @@
     "delay": "^4.3.0",
     "dirty-chai": "^2.0.1",
     "it-pair": "^1.0.0",
+    "libp2p": "libp2p/js-libp2p#chore/deprecate-old-peer-store-api",
```
vasco-santos, line 79: @jacobheun I am not sure if mocking the `peerStore` is the best thing to do here, or depend on `libp2p` as a dev dependency again. What do you think? It could be worth having a mocks repo for this kind of stuff, but we would need to make sure it was properly updated with new changes

src/rpc/handlers/get-providers.js
```diff
@@ -34,13 +34,7 @@ module.exports = (dht) => {
       dht._betterPeersToQuery(msg, peer)
     ])
 
-    const providers = peers.map((p) => {
-      if (dht.peerStore.has(p)) {
-        return dht.peerStore.get(p)
-      }
-
-      return dht.peerStore.put(new PeerInfo(p))
```
vasco-santos, line 42: Already known providers, no need to add to the peerStore. Moreover, the peerInfo is just created here, so there is nothing relevant to add

src/query/workerQueue.js
```diff
@@ -244,7 +244,6 @@ class WorkerQueue {
         if (this.dht._isSelf(closer.id)) {
           return
         }
-        closer = this.dht.peerStore.put(closer)
```
vasco-santos, line 247: No new information of the peer at this point

src/peer-routing/index.js
```diff
@@ -169,7 +173,7 @@ module.exports = (dht) => {
       result.paths.forEach((result) => {
         if (result.success) {
           success = true
-          dht.peerStore.put(result.peer)
+          dht.peerStore.addressBook.add(result.peer.id, result.peer.multiaddrs.toArray())
```
jacobheun, line 176: I like that this is now a discrete action. It's much clearer what's happening than a `put` imo.

src/query/workerQueue.js
jacobheun: Also it's unnecessary as the emit below will cause libp2p to update the peer store

package.json
vasco-santos: yeap, seems good!

PR #181, opened 2020-04-11
chore: use new content and peer routing apis

vasco-santos:
In the context of deprecating `peer-info` as described on [libp2p/js-libp2p#589](https://github.com/libp2p/js-libp2p/issues/589), this PR removes the `peer-info` from being used, as well as returned in the API, in favour of returning `{ id, multiaddrs }` in the same way as [ipfs](https://github.com/ipfs/js-ipfs/blob/447b44d1b64714f5ed0cafba166ad0a4dbbb587c/packages/ipfs/src/core/components/dht.js#L61) does. Moreover, it uses the new `topology`interface

BREAKING CHANGE: findProviders returns id and addrs properties instead of peer-info instance

This PR sits on top of #179 and #180, and those should be merged first and this PR rebased.

Needs:

- [x] #180 
- [x] [libp2p/js-libp2p#598](https://github.com/libp2p/js-libp2p/pull/598)
- [x]  [libp2p/js-libp2p-interfaces#42](https://github.com/libp2p/js-libp2p-interfaces/pull/42)
- [x]  [libp2p/js-libp2p-interfaces#43](https://github.com/libp2p/js-libp2p-interfaces/pull/43)


src/message/index.js
```diff
@@ -97,25 +97,18 @@ Message.TYPES = MESSAGE_TYPE
 Message.CONNECTION_TYPES = CONNECTION_TYPE
 
 function toPbPeer (peer) {
-  const res = {
+  return {
     id: peer.id.id,
-    addrs: peer.multiaddrs.toArray().map((m) => m.buffer)
+    addrs: (peer.multiaddrs || []).map((m) => m && m.buffer),
```
jacobheun, line 102: What's the reasoning for the check on m (`m && m.buffer`)? If it's possible for `m` to be null or undefined we should do a reduce, otherwise we'll end up with undefined in the `addrs` array for that value.

vasco-santos: It was me when trying to debug a bug to short circuit it. But it is not needed anymore 😆

PR #185, opened 2020-04-27
fix: add buffer

hugomrdias:


hugomrdias:
seems unrelated and PR run is green dunno whats happening

https://travis-ci.com/github/libp2p/js-libp2p-kad-dht/jobs/323739748#L323

vasco-santos:
I tried restarting the branch CI but it kept failing. So, I tried locally to run the tests, but I am getting the same result as in CI, while master branch is working for me locally. 

It is strange that the PR is getting green

vasco-santos:
Released this commit into `0.18.7` as latest and will also release this as `0.19.2`as a beta release

PR #193, opened 2020-05-28
chore: use libp2p0.28 release

vasco-santos:
Needs:

- [ ] final `js-libp2p@0.28` release


vasco-santos:
Duplicated of #195 closing

PR #194, opened 2020-06-04
fix: providers leaking resources on dht construction

vasco-santos:
After debugging one `js-ipfs` example hanging, I found this leaking issue.

When `js-libp2p` is created, the `dht` will be created if enabled. When the `dht` is enabled, it will be instantiated. During the instantiation, `Providers` was also instantiated and a setter was starting a timer. If the dht was created, but not started, when a process is closing, this timer was not cleared, since the stop was also not called.

### DIFF
--- a/src/providers.js
+++ b/src/providers.js
@@ -60,12 +60,9 @@ class Providers {
 
   /**
    * Start the provider cleanup service
+   * @returns {void}
    */
   start () {
-    if (this._cleaner) {
-      clearInterval(this._cleaner)
-    }
-
     this._cleaner = setInterval(
       () => this._cleanup(),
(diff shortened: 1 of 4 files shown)


src/providers.js
```diff
@@ -58,6 +58,20 @@ class Providers {
     this.syncQueue = new Queue({ concurrency: 1 })
   }
 
+  /**
+   * Start the provider cleanup service
+   */
+  start () {
+    if (this._cleaner) {
+      clearInterval(this._cleaner)
+    }
```
jacobheun, line 67: The check isn't necessary in `stop` and `start` since clearInterval ignores null/undefined values.

PR #199, opened 2020-07-08
fix: check for an existing connection before using the dialer

jacobheun:
Ideally libp2p should just get passed to the DHT and dial should be used, instead of interacting with the dialer. Currently the dialer does not check for existing connections, like libp2p.dial() does.

This issue was found at https://discuss.libp2p.io/t/random-walk-peer-discovery-adds-connections-to-same-peer/584/2. I tested the change against the sample code shared and it kept the connection count at 1 per peer.

We should probably look at updated the constructor in libp2p 0.29 to just take libp2p, instead of passing numerous components, and make that breaking change here to accompany it. We could also wait for the DHT updates before doing this.

### DIFF
--- a/src/network.js
+++ b/src/network.js
@@ -120,7 +120,11 @@ class Network {
     const id = to.toB58String()
     this._log('sending to: %s', id)
 
-    const conn = await this.dht.dialer.connectToPeer(to)
+    let conn = this.dht.registrar.connectionManager.get(to)
+    if (!conn) {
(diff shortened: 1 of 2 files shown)


vasco-santos:
I would wait for revamping the whole DHT to have that change. But it makes sense to do it!

vasco-santos:
Released in `0.19.8`

Strernd:
Thanks very much for the quick fix. Really love the work you're doing here.

PR #202, opened 2020-08-07
fix: replace node buffers with uint8arrays

achingbrain:
Replaces all use of node Buffers with Uint8Arrays

Depends on:

- [x] https://github.com/libp2p/js-libp2p-crypto/pull/180
- [x] https://github.com/multiformats/js-multiaddr/pull/140
- [x] https://github.com/libp2p/js-peer-id/pull/127
- [x] https://github.com/libp2p/js-libp2p-interfaces/pull/57

BREAKING CHANGES:

- Where node Buffers were returned, now Uint8Arrays are


test/query.spec.js
```diff
@@ -329,7 +329,7 @@ describe('Query', () => {
 
   it('disjoint path values', async () => {
     const peerId = dht.peerId
-    const values = ['v0', 'v1'].map(Buffer.from)
+    const values = ['v0', 'v1'].map((str) => uint8ArrayFromString(str))
```
Gozala, line 332: looks like you could pass `uint8ArrayFromString` directly to map, here and in two more cases below.

achingbrain: The signature for the function passed to `Array.map` is `function callback( currentValue[, index[, array]])`, so if you just pass `uint8ArrayFromString` it tries to interpret the `index` as a text encoding and fails.

HexaField:
https://github.com/ipfs/js-datastore-idb

This repo is still using Buffer

PR #204, opened 2020-11-04
feature: datastore content change events

HexaField:
This PR adds three events that provide state changes to the local datastore of the DHT only for content (ie non peer routing). When an entry is added, changed or removed. Includes tests. 

The intention here is to provide similar transparency to the state of DHT content that the connectionManager has with peer discovery/connection.

This solves https://github.com/libp2p/js-libp2p/issues/794

Usage: 

```javascript
dht.on('dht:add', (key, value, peerID) => { 
   console.log(`Added key ${key} with value ${value} from ${peerId}`) 
})
```

```javascript
dht.on('dht:changed', (key, oldValue, newValue, peerId) => { 
   console.log(`Changed key ${key} from ${oldValue} to ${newValue} from ${peerId}`) 
})
```

```javascript
dht.on('dht:removed', (key, value) => { 
   console.log(`Removed key ${key}`) 
})
```

I am open to changing the event names.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,20 @@
+<a name="0.20.1"></a>
+## [0.20.1](https://github.com/libp2p/js-libp2p-kad-dht/compare/v0.19.9...v0.20.1) (2020-08-11)
+
+
+### Bug Fixes
+
+* replace node buffers with uint8arrays ([#202](https://github.com/libp2p/js-libp2p-kad-dht/issues/202)) ([989be87](https://github.com/libp2p/js-libp2p-kad-dht/commit/989be87))
+
+
+### BREAKING CHANGES
+
+* - Where node Buffers were returned, now Uint8Arrays are
+
+* chore: remove gh dep urls
+
+
+
 <a name="0.20.0"></a>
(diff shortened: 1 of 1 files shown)


HexaField:
I've realised now it may make more sense to convert the values back to strings, since the current implementation outputs keys and values as buffer arrays.

vasco-santos (commented):
Hey @HexaField 
I have thought more about this solution. This will lead to decrease of performance due to extra datastore.get + deserialize for each put operation:

```
const rec = await dht.datastore.get(utils.bufferToKey(key))
currentCopy = Record.deserialize(rec)
```

I talked a bit with @jacobheun about this and we thought about having hooks for this.
For example, we can have a `onPut` function provided on the constructor (with a no operation function by default). Once put is done, this function would be called and the application layer can easily implement this type of features. 

This solution would not have any performance impact for regular users and would provide an easy way of people to add functionality based on state hooks. What do you think?

HexaField:
@vasco-santos Sounds good to me! Are you in any way against having an onDelete/onRemove under the same idea?

vasco-santos:
I think we can also have `onRemove` yes. It seems good to allow users to know when the DHT removed something from the datastore.

PR #205, opened 2020-12-01
Feature: onPut and onRemove events

HexaField:
This PR adds two events that provide notification of state changes to the local datastore of the DHT for content (ie non peer routing entries). Includes tests.

See https://github.com/libp2p/js-libp2p-kad-dht/pull/204 for previous discussion.

This solves libp2p/js-libp2p#794

Usage:

```javascript
dht.onPut = (record, peerId) => {}
dht.onRemove = (record) => {}
```

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -182,6 +182,9 @@ class KadDHT extends EventEmitter {
     this.contentFetching = contentFetching(this)
     this.contentRouting = contentRouting(this)
     this.peerRouting = peerRouting(this)
+
+    this.onPut = (record, peerId) => {}
+    this.onRemove = (record) => {}
   }
 
   /**
@@ -411,6 +414,7 @@ class KadDHT extends EventEmitter {
       utils.now() - record.timeReceived > c.MAX_RECORD_AGE) {
       // If record is bad delete it and return
       await this.datastore.delete(dsKey)
+      this.onRemove(record)
       return undefined
     }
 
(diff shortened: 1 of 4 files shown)


src/index.js
```diff
@@ -182,6 +182,9 @@ class KadDHT extends EventEmitter {
     this.contentFetching = contentFetching(this)
     this.contentRouting = contentRouting(this)
     this.peerRouting = peerRouting(this)
+
+    this.onPut = (record, peerId) => {}
```
vasco-santos, line 186: These should be provided via the constructor with a default to an empty function.

```js
const dht = new DHT({
  libp2p,
  // ... others
  onPut: (record, peerId) => {}, // User provided function
  onRemove: (record) => {} // User provided function
})
```

```js
  /**
   * Create a new KadDHT.
   * @param {Object} props
   * @param {Libp2p} [props.libp2p] the libp2p instance
   * @param {(record: Record, peerId: PeerId) => void} [onPut = () => {}]
   * @param {(record: Record) => void} [onRemove = () => {}]
   * ...
   */
   constructor({
     // ...
    onPut = () => {}
   }) {
     this.onPut = onPut
   }
```

vasco-santos (changes requested):
Thanks for the PR. Minor change needed on how to setup the functions.

vasco-santos (approved):
Great, thanks for the PR 👌

Issue #209, opened 2020-12-10
Buffer is not defined in browser


HexaField:
Trying to `put` a key into a DHT from the browser fails with the following error, despite all simulated aegir tests passing.

Seems to be an issue with either the webpack process or a dependency that is out of date. This may be the reason why this package was not pushed to the browser CDNs. I've tried tracking down the culprate to no avail.

```
Uncaught (in promise) ReferenceError: Buffer is not defined
    at n (index.min.js:2)
    at c.enqueue (index.min.js:2)
    at async e.exports.addPeerToQuery (index.min.js:2)
    at async Promise.all (:8080/index 0)
    at async e.exports.execute (index.min.js:2)
    at async Promise.all (:8080/index 0)
    at async e.exports.executePaths (index.min.js:2)
    at async e.exports.execute (index.min.js:2)
    at async Object.getClosestPeers (index.min.js:2)
    at async e.exports.getClosestPeers (index.min.js:2)
```

HexaField:
Further diagnosis shows this seems to be a problem with libp2p@0.29.0 (likely passing a bad function), updating to 0.29.4 seems to have fixed the issue.

HexaField:
The issue has returned without changing much. Not sure what to make of this.

jacobheun:
This is likely due to `xor-distance` using Buffer https://github.com/mafintosh/xor-distance/blob/v2.0.0/index.js#L5. Having another DHT peer connected would result in a query happening which would cause the peer-list logic to kick in, https://github.com/libp2p/js-libp2p-kad-dht/blob/87f85113c9078d1ca41858178d7bd66075c282ea/src/peer-list/peer-queue.js#L63.

Not sure why this wasn't caught during the switch over to Uint8Arrays, https://github.com/libp2p/js-libp2p-kad-dht/pull/202. @achingbrain any thoughts here on resolving this? Ideally we should keep the `allocUnsafe` usage in xor-distance in Nodejs as it's significantly more performant than Uint8Array usage. Perhaps we can find a workaround until we spend more time on DHT updates in the coming months.

HexaField:
I have forked, upgraded and tested `xor-distance` to use typed arrays https://github.com/HexaField/xor-distance-typedarray

Will now implement, test & make a PR.

achingbrain:
I've [PR'd](https://github.com/mafintosh/xor-distance/pull/3) `xor-distance` which should solve the issue.

[...]

PR #210, opened 2020-12-12
Fix/browser buffer to typedarray

HexaField:
Changes browser build to use `xor-distance-typedarray` package instead of `xor-distance` solving 'Buffer is not defined' errors

Has been tested externally to ensure browser environment conforms to changes

Solves https://github.com/libp2p/js-libp2p-kad-dht/issues/209


src/peer-list/peer-distance-list.js
```diff
@@ -1,6 +1,7 @@
 'use strict'
 
-const distance = require('xor-distance')
+const { isNode } = require('ipfs-utils/src/env')
+const distance = isNode ? require('xor-distance') : require('xor-distance-typedarray')
```
vasco-santos, line 4: Can we use the `browser` package json key to replace the module instead?

vasco-santos (changes requested):

HexaField:
Closing this as https://github.com/mafintosh/xor-distance/pull/3 will solve this once the repo is updated. Incidentally for the mean time for those who may be looking for a solution, importing https://github.com/feross/buffer with a quick hack `window.Buffer = window.buffer.Buffer` solves this problem for any problematic package.

PR #211, opened 2020-12-12
[Feature] adds removeLocal function

HexaField:
Adds the ability for a dht key/value pair entry to be removed from the local datastore. This essentially acts as dereferencing functionality for the peer.

### Usage

```js
await dht.removeLocal(uint8ArrayFromString('/v/hello'))
```

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -276,6 +276,7 @@ class KadDHT extends EventEmitter {
   /**
    * Remove the given key from the local datastore.
    * @param {Uint8Array} key
+   * @returns {Promise<void>}
    */
   async removeLocal (key) {
     this._log('removeLocal: %b', key)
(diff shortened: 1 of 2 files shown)


src/index.js
```diff
@@ -273,6 +273,24 @@ class KadDHT extends EventEmitter {
     return this.contentFetching.getMany(key, nvals, options)
   }
 
+  /**
+   * Remove the given key from the local datastore.
+   * @param {Uint8Array} key
```
vasco-santos, line 278:
```suggestion
   * @param {Uint8Array} key
   * returns {Promise<void>}
```

vasco-santos (changes requested):
Looks good, just a small nit with the jsdoc

vasco-santos (approved):
LGTM

PR #212, opened 2021-01-21
fix: do not throw on empty provider list

achingbrain:
If no content providers can be found, just be an empty iterator
instead of throwing.

This is necessary because multiple findProv strategies are tried
higher up the stack, at which point we make the decision whether
to throw or not.

This makes the behaviour consistent with the delegated content
routing module.

### DIFF
--- a/src/content-routing/index.js
+++ b/src/content-routing/index.js
@@ -148,10 +148,6 @@ module.exports = (dht) => {
         })
       })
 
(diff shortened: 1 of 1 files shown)


vasco-santos (approved):

vasco-santos:
🚢 `0.20.5`

Issue #183, opened 2020-04-23
Update the DHT to the latest libp2p spec


jacobheun:
The go-libp2p DHT has undergone a large effort to improve its performance. As a followup to its upcoming release a spec will be produced for other implementations to add support for. The spec is not yet complete, but this issue is being added to keep track of this effort.

Rough thoughts on timeline right now is that this issue could be unblocked for a Q3 or later implementation.

jacobheun:
Requires https://github.com/libp2p/specs/pull/108

p-shahi:
A related issue about spec compliance here: https://github.com/libp2p/js-libp2p-kad-dht/issues/408

@achingbrain This issue is quite old. Is this implementation in full compliance today? If not maybe we can itemize what's remaining into backlog items and determine priority.

achingbrain:
This implementation is spec compliant so I'm going to close this.

If any gaps are spotted, new issues should be opened.

PR #215, opened 2021-02-05
feat: periodically fill the routing table with KADIds

achingbrain:
In order to make queries in fewer hops it's useful to fill the routing
table with imaginary peers that have IDs that are uniformly distributed
over the sha2-256 value space.  This means we've always got a peer
ID on hand that's kind of xor-close to anything we might be looking up.

We can then issue period DHT queries for any actual existant peers that
are close to our imaginary peers to ensure we've got a good range of
peers we know about, so making queries requires fewer hops to locate
ones that are xor-close to the data we are interested in.

Related to https://github.com/libp2p/js-libp2p-kad-dht/issues/183

BREAKING CHANGE: `.start()` is now async and random walk has been removed


lidel:
Waiting for latest libp2p release.

BigLep:
@achingbrain : we removed "status/blocked" given llibp2p release has happened.

BigLep:
2021-06-14 note: this will get picked up by @achingbrain after Multiformats work.

src/index.js
```diff
@@ -237,17 +237,15 @@ class KadDHT extends EventEmitter {
...
-    }
-
     this._running = true
-    this.providers.start()
-    this._queryManager.start()
-    this.network.start()
 
-    // Start random walk, it will not run if it's disabled
-    this.randomWalk.start()
+    return Promise.all([
```
vasco-santos, line 242: Can you please add a breaking change message to the commit so that we do not forget it on merge?

package.json
```diff
@@ -54,6 +54,7 @@
     "heap": "~0.2.6",
     "interface-datastore": "^5.1.1",
     "it-first": "^1.0.4",
+    "it-length": "^1.0.3",
     "it-length-prefixed": "^5.0.2",
     "it-pipe": "^1.1.0",
     "k-bucket": "^5.1.0",
```
vasco-santos, line 60: we want to get rid of `xor-distance` below, right? 🎉

src/routing-table/index.js
```diff
@@ -0,0 +1,391 @@
+'use strict'
+
+// @ts-ignore
+const KBuck = require('k-bucket')
+const { xor: uint8ArrayXor } = require('uint8arrays/xor')
+const GENERATED_PREFIXES = require('./generated-prefix-list.json')
```
vasco-santos, line 6: we should have a browser empty list? I know we do not recommend using the DHT in the browser, but there are users who run it. It also will be part of the bundle, if we not have a browser alternative

test/generate-peers/generate-peer.go
```diff
@@ -0,0 +1,85 @@
+package main
```
vasco-santos, line 1: I wonder if we really want this in the tests, or a top level folder

vasco-santos (commented):
This is looking good, way more clear routing-table!

With the periodically fill of the routing table, I think we can just drop the random walk? What do you think? From what I remember, go also dropped it

src/routing-table/index.js
achingbrain: If they use it in the browser and we've removed random walk, it probably shouldn't be an empty list otherwise peer discovery will not work properly.  We could always have a 'light' version that's like, 1/4 of the values or something?

src/index.js
achingbrain: Done!

package.json
achingbrain: Done!

src/routing-table/index.js
vasco-santos: yeah, good point. Ok, I think we can start with a 1/4 of the list for now. These are the kind of things I wish JS testground had shipped. 

We will likely need to revisit, but we will need to revisit all browser story either way. So, yes, let's do 1/4 for now.

achingbrain: Done!

test/generate-peers/generate-peer.go
achingbrain: I think it's useful to show we're working with the same data

vasco-santos (approved):
LGTM

PR #217, opened 2021-04-13
chore: update deps

vasco-santos:
Updates dependencies, switch from protons to protobufjs as the rest of the stack

`aegir` since 31+ does not close the test process if there are outstanding resources being used. This is great as we are able to catch new errors easily. There is a known error where queries are not properly closed which means the tests will not be closed in the end. I am deferring this fix to a follow up PR (need to check if we have an issue, otherwise create one)

Needs:

- [x] https://github.com/libp2p/js-libp2p/pull/904
- [x] libp2p final release


package.json
```diff
@@ -68,31 +66,31 @@
...
     "xor-distance": "^2.0.0"
   },
   "devDependencies": {
     "@types/debug": "^4.1.5",
     "aegir": "^30.3.0",
     "async-iterator-all": "^1.0.0",
-    "chai": "^4.2.0",
+    "chai": "^4.3.0",
     "chai-checkmark": "^1.0.1",
     "datastore-level": "^4.0.0",
```
achingbrain, line 81:
```suggestion
    "datastore-level": "^5.0.0",
```

package.json
```diff
@@ -53,13 +51,13 @@
     "heap": "~0.2.6",
     "interface-datastore": "^3.0.4",
```
achingbrain, line 52:
```suggestion
    "interface-datastore": "^4.0.0",
```

achingbrain (approved):

package.json
```diff
@@ -6,9 +6,11 @@
   "main": "src/index.js",
   "scripts": {
     "lint": "aegir lint",
+    "build": "npm run build:proto && npm run build:proto-types && aegir build",
```
achingbrain, line 9: A prepare script to allow branches of this module to be used as gh url deps:

```suggestion
    "prepare": "npm run build",
    "build": "npm run build:proto && npm run build:proto-types && aegir build",
```

package.json
```diff
@@ -6,9 +6,12 @@
   "main": "src/index.js",
   "scripts": {
     "lint": "aegir lint",
+    "prepare": "npm run build",
+    "build": "npm run build:proto && npm run build:proto-types && aegir build",
+    "build:proto": "pbjs -t static-module -w commonjs --force-number --no-verify --no-delimited --no-create --no-beautify --no-defaults --lint eslint-disable -o src/message/dht.js ./src/message/dht.proto",
```
achingbrain, line 11:
```suggestion
    "build:proto": "pbjs -t static-module -w commonjs -r libp2p-kad-dht --force-number --no-verify --no-delimited --no-create --no-beautify --no-defaults --lint eslint-disable -o src/message/dht.js ./src/message/dht.proto",
```

achingbrain (approved):

vasco-santos:
Prior to https://github.com/libp2p/js-libp2p/commit/b4fb9b7bf266ba03c4462c0a41b1c2691e4e88d4 the event was emitted with empty addresses, which meant the findPeer was working, but no addresses existed. In addition, test dht connect now also adds to the addressBook the peer addresses onConnect, as libp2p would do if we were using its dialer.

achingbrain (approved):

PR #218, opened 2021-04-20
fix: add pbjs namespace

achingbrain:
To prevent namespace conflicts with protobuf message names.

Refs https://github.com/ipfs/js-ipfs-bitswap/pull/323


achingbrain:
Fixed in #217

PR #220, opened 2021-05-15
chore: update to new multiformats

achingbrain:
Uses new CID class.  Also swaps out various chai deps for version
bundled with aegir, updates aegir and fixes tests that start services
without tearing them down which stops the tests from finishing.


src/utils.js
```diff
@@ -204,9 +204,13 @@ exports.withTimeout = (asyncFn, time) => {
     return Promise.race([
       asyncFn(...args),
       new Promise((resolve, reject) => {
-        setTimeout(() => {
+        const timeout = setTimeout(() => {
           reject(errcode(new Error('Async function did not complete before timeout'), 'ETIMEDOUT'))
         }, time)
+
+        if (timeout && timeout.unref) {
```
vasco-santos, line 211: `setTimeout` return a number, `timeoutId`. Type checker is failing here as a consequence

src/content-routing/index.js
```diff
@@ -10,7 +10,7 @@ const Query = require('../query')
 const utils = require('../utils')
 
 /**
- * @typedef {import('cids')} CID
+ * @typedef {import('multiformats').CID} CID
```
vasco-santos, line 13:
```suggestion
 * @typedef {import('multiformats/cid').CID} CID
```

src/message/dht.d.ts
```diff
@@ -28,19 +28,34 @@ export class Record implements IRecord {
     constructor(p?: IRecord);
 
     /** Record key. */
-    public key: Uint8Array;
+    public key?: (Uint8Array|null);
```
vasco-santos, line 31: was this generated?

src/providers.js
```diff
@@ -10,7 +10,7 @@ const c = require('./constants')
 const utils = require('./utils')
 
 /**
- * @typedef {import('cids')} CID
+ * @typedef {import('multiformats').CID} CID
```
vasco-santos, line 13:
```suggestion
 * @typedef {import('multiformats/cid').CID} CID
```

src/rpc/handlers/add-provider.js
```diff
@@ -1,6 +1,6 @@
 'use strict'
 
-const CID = require('cids')
+const { CID } = require('multiformats')
```
vasco-santos, line 3:
```suggestion
const { CID } = require('multiformats/cid')
```

src/rpc/handlers/get-providers.js
```diff
@@ -1,6 +1,6 @@
 'use strict'
 
-const CID = require('cids')
+const { CID } = require('multiformats')
```
vasco-santos, line 3:
```suggestion
const { CID } = require('multiformats/cid')
```

src/index.js
```diff
@@ -254,12 +258,12 @@ class KadDHT extends EventEmitter {
    *
    * @returns {Promise<void>}
    */
-  stop () {
+  async stop () {
     this._running = false
     this.randomWalk.stop()
-    this.providers.stop()
+    await this.network.stop()
```
vasco-santos, line 264: This does not need to be a promise, but `network.stop` and `network.start` need to be fixed. Registrar does not return a promise anymore: https://github.com/libp2p/js-libp2p/blob/master/src/registrar.js#L87

test/providers.spec.js
```diff
@@ -1,15 +1,13 @@
 /* eslint-env mocha */
 'use strict'
 
-const chai = require('chai')
-chai.use(require('dirty-chai'))
-const expect = chai.expect
+const { expect } = require('aegir/utils/chai')
 const { MemoryDatastore } = require('interface-datastore')
-const CID = require('cids')
+const { CID } = require('multiformats')
```
vasco-santos, line 6:
```suggestion
const { CID } = require('multiformats/cid')
```

test/utils/create-values.js
```diff
@@ -1,16 +1,16 @@
 'use strict'
 
-const multihashing = require('multihashing-async')
-const CID = require('cids')
+const { CID } = require('multiformats')
```
vasco-santos, line 3:
```suggestion
const { CID } = require('multiformats/cid')
```

vasco-santos (changes requested):
Mostly minor stuff, some are learnings from @rvagg cid import flags in ipfs-repo

src/utils.js
achingbrain: In node it returns a [Timeout](https://nodejs.org/api/timers.html#timers_class_timeout) which has an `unref` method property.  Weird, maybe we need to tell ts that it's used under node.

src/message/dht.d.ts
achingbrain: Yes, I'm not sure why this changed, new patch release of protobufjs I guess?

src/index.js
achingbrain: I've removed all of the unnecessary async.

src/utils.js
achingbrain: I'm managed to refactor this away after removing the unnecessary async and using the `p-timeout` module instead of `Promise.race`.

achingbrain:
I've marked this as a draft so we can merge it when #3556 is green as it'll bubble everything all the way up.  Please do continue review and/or approve as appropriate.

src/index.js
```diff
@@ -29,7 +29,7 @@ const Record = libp2pRecord.Record
  * @typedef {import('interface-datastore').Datastore} Datastore
  * @typedef {import('libp2p/src/dialer')} Dialer
  * @typedef {import('libp2p/src/registrar')} Registrar
- * @typedef {import('cids')} CID
+ * @typedef {import('multiformats').CID} CID
```
vasco-santos, line 32:
```suggestion
 * @typedef {import('multiformats/cid').CID} CID
```

vasco-santos (changes requested):

PR #229, opened 2021-09-27
feat: ping old DHT peers before eviction

achingbrain:
Updates the DHT contact logic to ping the old peers when buckets are full and only evict if the peer does not respond.

Runs ping requests in a queue to prevent flooding the network if lots and lots of peers are added very quickly.

Uses the generic libp2p ping, rather than the DHT RPC ping, I'm not sure if this is correct.  It looks like `go-libp2p-kad-dht` [just tries to dial the remote peer](https://github.com/libp2p/go-libp2p-kad-dht/blob/35cca9a2555a0f149b4b9ad2e32555f9dc18ec37/rtrefresh/rt_refresh_manager.go#L173-L190) and treats that as success, so maybe this should just do that?

### DIFF
--- a/src/routing-table/index.js
+++ b/src/routing-table/index.js
@@ -14,6 +14,8 @@ const log = Object.assign(debug('libp2p:dht:routing-table'), {
 })
 // @ts-ignore
 const length = require('it-length')
+const { default: Queue } = require('p-queue')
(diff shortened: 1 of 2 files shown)


achingbrain:
Going to merge this as further refactoring is necessary as part of #233

PR #258, opened 2021-12-01
feat: remove dht.getPublicKey

achingbrain:
The `dht.getPublicKey` method is a specialised `dht.get` so you should just call that with the right key value.

This lets us remove the `libp2p-crypto` dependency as we were using it to unmarshal the returned key - if you are interested in a node's public key you're probably about to do something with it in which case you probably have a dep on `libp2p-crypto` so no need for this module to have one.

BREAKING CHANGE: `dht.getPublicKey` has been removed


src/dual-kad-dht.js
```diff
@@ -333,7 +334,12 @@ class DualKadDHT extends EventEmitter {
       throw errCode(new Error('Failed to load public key'), 'ERR_FAILED_TO_LOAD_KEY')
     }
 
-    const peerId = new PeerId(peer.id, undefined, pk)
+    const peerId = await PeerId.createFromPubKey(pk)
+
+    if (peerId.toB58String() !== peer.toB58String()) {
```
achingbrain, line 339: We only create the PeerId from the public key to add it to the key book.  Do we even want to be doing that at this level?

achingbrain: AFAIK it's only used when you're dealing with a peer with an RSA key (not Ed25519) and they should become less common over time?

hugomrdias:
Im being picky here but that stablelib doesnt do the node/browser swap includes both and if the bundler doesnt do treeshaking it includes even more stuff. 

Perfect scenario would be `libp2p-crypto/random` or a new package with the https://github.com/hugomrdias/iso-random-stream/blob/master/src/random.browser.js (could be improved with the clever `for` from the stablelib impl) , https://github.com/hugomrdias/iso-random-stream/blob/master/src/random.js and the browser swap in package.json.

the rest LGTM

achingbrain:
> Im being picky here

That's fine, I've not got a strong opinion about which `randomBytes` implementation to use.

> Perfect scenario would be `libp2p-crypto/random` 

I disagree because this module only needs a random number generator, not a full crypto suite.

> or a new package with the hugomrdias/iso-random-stream@master/src/random.browser.js (could be improved with the clever for from the stablelib impl) , hugomrdias/iso-random-stream@master/src/random.js and the browser swap in package.json.

Sounds preferable - would merge!

achingbrain:
@hugomrdias I've reverted to `iso-random-stream` but it needs https://github.com/hugomrdias/iso-random-stream/pull/87 before the build will pass here.

achingbrain:
Superseded by #300

PR #272, opened 2021-12-29
feat: async peer store

achingbrain:
Refactors interfaces and classes used by `libp2p-interfaces` to use the async peer store from https://github.com/libp2p/js-libp2p/pull/1058

BREAKING CHANGE: peerstore methods are now all async

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,12 +60,12 @@ import { create } from 'libp2p-kad-dht'
 /**
  * @param {Libp2p} libp2p
  */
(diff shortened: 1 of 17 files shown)


vasco-santos (commented):
This looks good! There is an issue with the types somewhere though

achingbrain:
Yeah - it needs https://github.com/libp2p/js-libp2p-interfaces/pull/128 to land first

PR #300, opened 2022-02-21
feat: convert to typescript

achingbrain:
- Converts to typescript
- Only named exports
- No more CJS, only ESM
- Runs tests on all supported environments
- Adds auto-publish
- Adds dependabot

BREAKING CHANGE: switch to named exports, ESM only


wemeetagain (approved):
Reviewed for sanity, this is too big to really do a deep review.

PR #316, opened 2022-04-17
fix: max listeners warning

achingbrain:
The query timeout controller signal gets passed to all of the network operations etc that run during a query so can end up with lots of listeners for it's 'abort' event, which can trigger `MaxListenersExceededWarning` in node which warns of a memory leak, even though there is no leak.


src/query/manager.ts
```diff
@@ -89,6 +89,14 @@ export class QueryManager implements Startable, Initializable {
       // don't let queries run forever
       timeoutController = new TimeoutController(DEFAULT_QUERY_TIMEOUT)
       options.signal = timeoutController.signal
+
+      // this signal will get listened to for network request, etc
```
achingbrain, line 93:
```suggestion
      // this signal will get listened to for network requests, etc
```

PR #328, opened 2022-05-03
Export ./content-routing, ./kad-dht and ./message

ukstv:
When working on lower-level DHT provider I noticed that I can not directly get types and values from content-routing, kad-dht and message-related files. This PR makes them available for an external package to use.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,6 +29,12 @@
   "exports": {
     ".": {
       "import": "./dist/src/index.js"
+    },
+    "./content-routing": {
+      "import": "./dist/src/content-routing/index.js"
+    },
+    "./message": {
+      "import": "./dist/src/message/index.js"
     }
   },
   "eslintConfig": {
--- a/package.json
+++ b/package.json
@@ -35,6 +35,9 @@
     },
     "./message": {
       "import": "./dist/src/message/index.js"
+    },
+    "./kad-dht": {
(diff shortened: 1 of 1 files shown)


achingbrain:
Could you go into detail a bit more about what you are trying to do and how this solves it?

I'm reluctant to expose internal files unless absolutely necessary as they then become part of the public API.

Really everything should be contained within the [DHT interface](https://github.com/libp2p/js-libp2p-interfaces/blob/master/packages/libp2p-interfaces/src/dht/index.ts#L135-L180).

ukstv:
The goal is simple: provide a record via DHT. The general algorithm as implemented in [`ContentRouting::provide`](https://github.com/libp2p/js-libp2p-kad-dht/blob/master/src/content-routing/index.ts#L64) is (1) find closest peers, (2) send them `ADD_PROVIDER` message. If I am going through [dht.addProvide](https://github.com/libp2p/js-libp2p-kad-dht/blob/master/src/content-routing/index.ts#L64), it does both steps, as intended.

What I would like to do though, is to not use JS DHT for finding closest peers. I find the closest peers through other means, that is using a client connection to a go-ipfs node. The expectation is that go-ipfs peer-finding is faster. To do that, I have to essentially reimplement ContentRouting::provide in my code. And this requires access to the components exposed in this PR. If they are not exposed, I have to copy the code, which I would like to avoid.

PR #335, opened 2022-05-14
fix: remove p-map dependency

abetaev:
this module depends on node API making @libp2p/kad-dht unusable in browser.

### DIFF
--- a/package.json
+++ b/package.json
@@ -159,7 +159,6 @@
     "k-bucket": "^5.1.0",
     "multiformats": "^9.6.3",
     "p-defer": "^4.0.0",
-    "p-map": "^5.3.0",
     "p-queue": "^7.2.0",
     "private-ip": "^2.3.3",
     "protons-runtime": "^1.0.4",
--- a/src/peer-list/peer-distance-list.ts
+++ b/src/peer-list/peer-distance-list.ts
@@ -1,5 +1,4 @@
 import * as utils from '../utils.js'
-import pMap from 'p-map'
 import { compare as uint8ArrayCompare } from 'uint8arrays/compare'
 import { xor as uint8ArrayXor } from 'uint8arrays/xor'
 import type { PeerId } from '@libp2p/interfaces/peer-id'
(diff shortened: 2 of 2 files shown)


wemeetagain (approved):
LGTM

achingbrain:
> making @libp2p/kad-dht unusable in browser

This module is used in the browser by libp2p and ipfs, tested in CI, etc so this may not solve your problem.  What's the error message you are seeing?

abetaev:
@achingbrain 

> this may not solve your problem

actually, it did,

>  What's the error message you are seeing?

`p-map -> aggregate-error -> clean-stack` - there's [import of `os` module](https://github.com/sindresorhus/clean-stack/blob/main/index.js#L1).
particularly in firefox i saw this stacktrace before change:
```
Uncaught Error: Module "os" has been externalized for browser compatibility and cannot be accessed in client code.
    get os:3
    <anonymous> index.js:6
[os:3:10](browser-external:os)
    get os:3
    <anonymous> index.js:6
    InnerModuleEvaluation self-hosted:2331
    InnerModuleEvaluation self-hosted:2331
    evaluation self-hosted:2292
```

> This module is used in the browser by libp2p and ipfs, tested in CI

i did not look into browser tests, but i remember having a thought of surprise about that. all browser tests pass in libp2p and in this package, but i am getting errors in the real browser.

achingbrain:
In `clean-stack` the os import is [replaced with `false`](https://github.com/sindresorhus/clean-stack/blob/main/package.json#L42) by the `browser` key in `package.json` - how ever you are consuming the code from this module may not be respecting browser overridess.

abetaev:
@achingbrain you are right, i did a little digging...

[vite replaces the actual import with that exception while in dev mode](https://github.com/vitejs/vite/blob/79a8c852e84506a844057b0a2fcf9448fb8123e1/packages/vite/src/node/plugins/resolve.ts#L336). in production mode it replaces it with an empty object.

it is, probably, a matter of taste, but the way vite handles "externalized" modules seems fair to me: if not treated properly in code, they might cause undefined behavior.

Issue #324, opened 2022-05-02
README.md references dead URLs


zeim839:
[README.md](https://github.com/libp2p/js-libp2p-kad-dht/blob/master/README.md): Github is not rendering the _Peer Routing_, _Content Routing_ and _Peer Discovery_ sections. Visiting the referenced URLs returns a `404: Not Found` error.

achingbrain:
Great spot, thanks so much - would you like to open a PR that fixes the docs?

zeim839:
Apologies, I should've done that in the first place but I hadn't looked for the updated image src. at the time. I'll submit a PR in a few hours, thanks for reaching out.

PR #342, opened 2022-05-24
docs: fix readme.md

zeim839:
Solves issue #324 

**Issues addressed:**
1. Dead Peer Routing, Content Routing, Peer discovery badge images. 
2. Dead/incorrect `Open an issue...` link. This now points to the js-libp2p-kad-dht issues page, instead of a dead IPFS issues page. I also felt that pointing to the local issues page rather than IPFS was more reasonable (this is what other libp2p repos tend to do). 
3. Updated contributing badge to point to new URL. 

For the peer routing, etc. badges, I've linked the images to their respective deprecated and read-only repositories - this should prevent any future 404s and is what the updated libp2p-interfaces repo does. Clicking on the images will still redirect to the updated repositories. 

Please note that there's still a dead URL in the `dependency status` badge (below the title header). The entire domain is down, so I wasn't sure what to replace it with. 

That said, all other URL's/badges should now be working and updated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,7 +91,7 @@ Feel free to join in. All welcome. Open an [issue](https://github.com/libp2p/js-
 
 This repository falls under the IPFS [Code of Conduct](https://github.com/ipfs/community/blob/master/code-of-conduct.md).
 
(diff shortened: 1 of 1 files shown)


wemeetagain (approved):
:heart:

achingbrain (approved):

PR #344, opened 2022-06-01
fix: ping peers with correct protocol

achingbrain:
Fixes a bug in the routing table peer eviction routine whereby the wrong protocol was being used.


BigLep:
@achingbrain : do we have a test for this change somewhere?

Issue #89, opened 2019-03-18
Enable browser tests


vasco-santos:
Currently, this module is not running tests in the browser, as a result of the usage of `net.createServer`. We should add a hook for it or separate node tests from browser tests.

PR #352, opened 2022-06-30
fix: enable and fix browser tests

abetaev:
[related issue](https://github.com/libp2p/js-libp2p-kad-dht/issues/89)

i wasn't able to figure out why `rpc/index.ts` test fails in browser


src/query-self.ts
```diff
@@ -1,4 +1,4 @@
-import { setMaxListeners } from 'events'
+import { setMaxListeners } from './utils.js'
```
mpetrunic, line 1: why? events should work in browser

abetaev: maybe it will with [this package](https://www.npmjs.com/package/events) in dependencies.

should i update the PR?

mpetrunic: yeah, lets try that instead

abetaev: i have updated. 

the `npm run test:node`runs tests ending with `.node.ts`, but `npm test` does not. is it the way it supposed to work?

mpetrunic: Are you sure, they are both same command?
```
    "test": "aegir test -t node",
    "test:node": "aegir test -t node",
```

mpetrunic (approved):

wemeetagain (approved):
thanks @abetaev

PR #371, opened 2022-09-07
fix: re-enable ensuring queries run along disjoint paths

zeroxbt:
I noticed that during the getClosestPeers operation, the query paths might overlap and the same nodes will be queried multiple times.
There is no need to query a node if it has been queried during the execution of another query path, so the `peersSeen` set should be shared among different query paths belonging to the same query.


wemeetagain (approved):
LGTM

@achingbrain should also have a look

achingbrain:
This feature is called disjoint paths - it's in the S/Kademlia paper and this impl used to support it.

Notably the go KAD-DHT implementation does not have this feature which is why it was removed from this repo, though I notice the rust implementation [does support it](https://github.com/libp2p/rust-libp2p/pull/1473) so there's no harm in re-adding it here, and is certainly worth doing if it makes the queries run faster.

Thanks for opening this!

src/peer-routing/index.ts
```diff
@@ -235,8 +235,6 @@ export class PeerRouting implements Initializable {
     }
 
     for await (const event of this.queryManager.run(key, tablePeers, getCloserPeersQuery, options)) {
-      yield event
```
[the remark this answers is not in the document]
zeroxbt: I would expect [this function](https://github.com/libp2p/js-libp2p/blob/111e75d05e41eb2141fefb2cfafee0937fe068c2/src/peer-routing.ts#L171) to yield the `kBucketSize` sorted peers, but it's yielding a lot more than expected (not all are `FINAL_PEER`), and not necessarily sorted (because it merges events yielded by both `lan` and `wan` dhts).

src/query/manager.ts
```diff
@@ -140,7 +140,7 @@ export class QueryManager implements Startable, Initializable {
       }
 
       // make sure we don't get trapped in a loop
-      const peersSeen = new Set()
+      const peersSeen = new Set<string>()
```
achingbrain, line 143: You can also use `PeerSet` from [@libp2p/peer-collections](https://www.npmjs.com/package/@libp2p/peer-collections) for this.

achingbrain:
The final thing to do to get this in is to write some tests.

For an example you can see how we create predictable network topologies in [test/query.spec.js](https://github.com/libp2p/js-libp2p-kad-dht/blob/master/test/query.spec.ts#L382-L428) and run a query which records the peers it passed through to get to the final value.

You should be able to validate in a test that queries stop when passing through the same node twice.

test/query.spec.ts
```diff
@@ -427,6 +427,40 @@ describe('QueryManager', () => {
...
+      .map(event => {
+        if (event.type !== EventTypes.PEER_RESPONSE && event.type !== EventTypes.VALUE) {
+          throw new Error(`Unexpected query event type ${event.type}`)
+        }
+
+        return event.from
+      })
+
+      console.log(traversedPeers.map(e=> e.toString()))
+
```
achingbrain, line 458:
```suggestion
```

test/query.spec.ts
```diff
@@ -427,6 +427,40 @@ describe('QueryManager', () => {
...
+    expect(traversedPeers).lengthOf(7)
```
achingbrain, line 459:
```suggestion
      })

    expect(traversedPeers).lengthOf(7)
```

src/query/manager.ts
```diff
@@ -139,6 +140,9 @@ export class QueryManager implements Startable, Initializable {
         return
       }
 
+      // make sure we don't get trapped in a loop
+      const peersSeen = new PeerSet
```
achingbrain, line 144:
```suggestion
      const peersSeen = new PeerSet()
```

achingbrain (approved):

achingbrain:
Thanks!

PR #375, opened 2022-09-29
feat: tag kad-close peers

achingbrain:
When peers are added or removed from the kbuckets, if the `kBucketSize` closest peers change, update their peer store tags to ensure we don't close connections to them when the connection manager reaches it's max connections limit.


mpetrunic (approved):
LGTM, @wemeetagain  do you also wanna review this?

achingbrain:
I thinking maybe we want to tag them with `keep-alive` too, this will mean we try to reconnect on node startup.

mpetrunic:
> I thinking maybe we want to tag them with `keep-alive` too, this will mean we try to reconnect on node startup.

I was contemplating the same but I feel like it's better if we tag them differently, on startup or pruning we can determine priorities. Like we could prune 'kad-close' tags before 'keep-alive' tags or something.

Or did you mean putting two tags?

achingbrain:
I was thinking `keep-alive` in addition to `kad-close`.

PR #380, opened 2022-10-07
add protocols to messages

zeroxbt:
After calling `getClosestPeers`, I was expecting that the protobook would be updated with protocols supported by newly discovered peers, but I noticed that only the addressbook is updated.

### DIFF
--- a/package.json
+++ b/package.json
@@ -184,7 +184,7 @@
     "varint": "^6.0.0"
   },
   "devDependencies": {
-    "@libp2p/interface-mocks": "^4.0.3",
+    "@libp2p/interface-mocks": "^6.0.0",
     "@libp2p/peer-id-factory": "^1.0.18",
     "@libp2p/peer-store": "^3.1.4",
     "@types/lodash.random": "^3.2.6",
--- a/package.json
+++ b/package.json
@@ -137,7 +137,7 @@
     "release": "aegir release"
   },
   "dependencies": {
-    "@libp2p/components": "^3.0.1",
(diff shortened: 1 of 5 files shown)


achingbrain:
Supported protocols are normally exchanged by running the [identify protocol](https://github.com/libp2p/specs/tree/master/identify) after dialing a peer, not gathered from DHT query results.  They can also change fairly dynamically so may not be valid for very long.

That said this change could be useful, but it really needs to start with a [spec](https://github.com/libp2p/specs/tree/master/kad-dht) change first so other implementations behave in the same way.

Can you please open an issue in the [spec repo](https://github.com/libp2p/specs/tree/master/kad-dht) with the reasoning for discussion.

The `protocols` field will need to be added to the [rpc protobuf](https://github.com/libp2p/specs/tree/master/kad-dht#rpc-messages).

zeroxbt:
Thanks for replying @achingbrain . 
For more context, this feature is useful when you need to perform some protocol operation on the closest peers. To achieve this, after calling getClosestPeers, you'd need to filter peers that don't support the protocol before dialing them. 
The problem is that with the current implementation it's not always possible to filter newly discovered peers by protocols since they are not added to the protobook on discovery.

achingbrain:
Closing because this module is an implementation of a spec, and this feature isn't part of the spec.

To service your use case you can search for close peers, then just use `libp2p.dialProtocol` - if they support the protocol then all good, if not then handle the failure case.

PR #404, opened 2022-11-29
feat: allow passing ProvidersInit in KadDHT constructor

joeltg:
The `Providers` component takes a `ProvidersInit` object, but there's not actually a way to provide it with one from the `KadDHT` constructor. This means users aren't able to override the defaults for how long provider records are valid for and how frequently to clean up expired provider records.

Feel free to close this if there's a different way you'd rather do it, or if for some reason it's not allowed on purpose (though I can't imagine why), but it's such a small change I thought I'd just open it as a PR.

Thanks for your time; this would really help me out!

### DIFF
--- a/src/kad-dht.ts
+++ b/src/kad-dht.ts
@@ -82,7 +82,8 @@ export class KadDHT extends EventEmitter<PeerDiscoveryEvents> implements DHT {
       pingTimeout,
       pingConcurrency,
       maxInboundStreams,
-      maxOutboundStreams
+      maxOutboundStreams,
+      providers: providersInit,
     } = init
 
     this.running = false
@@ -102,7 +103,7 @@ export class KadDHT extends EventEmitter<PeerDiscoveryEvents> implements DHT {
       protocol: this.protocol
     })
 
-    this.providers = new Providers(components)
+    this.providers = new Providers(components, providersInit ?? {})
 
     this.validators = {
       ...recordValidators,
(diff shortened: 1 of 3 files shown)


achingbrain (changes requested):
LGTM, could you please add a test to ensure we are actually passing the config object to the providers constructor

src/kad-dht.ts
```diff
@@ -82,7 +82,8 @@ export class KadDHT extends EventEmitter<PeerDiscoveryEvents> implements DHT {
       pingTimeout,
       pingConcurrency,
       maxInboundStreams,
-      maxOutboundStreams
+      maxOutboundStreams,
+      providers: providersInit,
```
achingbrain, line 86:
```suggestion
      providers: providersInit
```

test/kad-dht.spec.ts
```diff
@@ -91,6 +91,21 @@ describe('KadDHT', () => {
       expect(dht).to.have.property('getMode')
       expect(dht).to.have.property('setMode')
     })
+
+    it('forward providers init options to providers component', async () => {
+      const dht = await tdht.spawn({
+        kBucketSize: 5,
+        providers: {
+          cleanupInterval: 60,
+          provideValidity: 60 * 10,
```
achingbrain, line 100:
```suggestion
          provideValidity: 60 * 10
```

achingbrain (approved):

test/kad-dht.spec.ts
```diff
@@ -91,6 +91,21 @@ describe('KadDHT', () => {
...
+
+    it('forward providers init options to providers component', async () => {
+      const dht = await tdht.spawn({
+        kBucketSize: 5,
+        providers: {
+          cleanupInterval: 60,
+          provideValidity: 60 * 10
+        }
+      })
+
```
achingbrain, line 103:
```suggestion

```

achingbrain:
Thanks!

Issue #381, opened 2022-10-07
Inconsistency between message keys in ADD_PROVIDER and GET_PROVIDER (cid bytes vs multihash bytes)


joeltg:
In src/content-routing/index.ts, [`ContentRouting.provide` uses `key.bytes` as the value for `Message.key`](https://github.com/libp2p/js-libp2p-kad-dht/blob/708f19b7d71e3488cdcd8a4b81ae4143431f5073/src/content-routing/index.ts#L70), while [`ContentRouting.findProviders` uses _`key.multihash.bytes`_](https://github.com/libp2p/js-libp2p-kad-dht/blob/708f19b7d71e3488cdcd8a4b81ae4143431f5073/src/content-routing/index.ts#L132).

I found this while debugging a minimal local DHT in which all my provides were succeeding but all attempts to look up those providers failed with `Error: stream reset` on the client peer (the peer calling findProviders) and **failing silently** on the host (the peer with the provider record) - just `incoming GET_PROVIDERS from $PEER_ID` in the logs and then nothing.

It turns out that [`CID.decode` in src/rpc/handlers/get-providers.ts](https://github.com/libp2p/js-libp2p-kad-dht/blob/680b90a4f3b793fff16dfa1c2450faaacc2b3383/src/rpc/handlers/get-providers.ts#L45) was throwing an error (`Incorrect length`) because the key of the incoming GET_PROVIDERS message was just the multihash bytes! This happens to work for CIDv0s but not for CIDv1s.

joeltg:
Ah - I see both go-libp2p and js-libp2p recently switched from CIDs to multihashes across-the-board for DHT queries (https://github.com/libp2p/js-libp2p-kad-dht/pull/237). I guess just the `CID.decode` part in `GetProvidersHandler.handle` got missed?

achingbrain:
This is a bug, we should be running queries on the multihash bytes and not the full CID since two CIDs could share a multihash.  Would you like to submit a quick PR with a fix and a test?

PR #405, opened 2022-11-29
fix: use multihash bytes for provide messages keys

joeltg:
This fixes a bug where `ADD_PROVIDER` message keys were CID bytes instead of multihash bytes, and changes the CIDs generated in tests to be V1s with `raw` codec instead of V0s so that the difference is actually tested..


joeltg:
(fixes #381)

achingbrain (approved):

PR #420, opened 2023-02-09
chore: Upgrade `aegir` to 38.1.2

maschad:


src/query/query-path.ts
```diff
@@ -98,7 +97,7 @@ export async function * queryPath (options: QueryPathOptions) {
    * Adds the passed peer to the query queue if it's not us and no
    * other path has passed through this peer
    */
-  function queryPeer (peer: PeerId, peerKadId: Uint8Array) {
+  function queryPeer (peer: PeerId, peerKadId: Uint8Array): void {
```
achingbrain, line 100: The changes to this function break some of the tests.  It looks like you're trying to clean up the TODOs about `p-queue@7` - thanks!  It's worth doing that so could you please open a separate PR with just the changes to this method then it should be easier to diagnose the breakage.

maschad: Good point! Will do. I opened an issue https://github.com/libp2p/js-libp2p-kad-dht/issues/429

PR #464, opened 2023-04-27
feat: use symbols to return routers

achingbrain:
The DHT interface is similar to the content/peer routing interfaces but different.

In order to detect content/peer routers and peer discovery implementations in [arbitrary libp2p modules](https://github.com/libp2p/js-libp2p/pull/1563) use the exported symbols to create getters for them.

The DHTContentRouting and DHTPeerRouting classes have been ported over from libp2p.


src/index.ts
```diff
@@ -1,14 +1,22 @@
...
-import type { Selectors, Validators } from '@libp2p/interface-dht'
 import type { Registrar } from '@libp2p/interface-registrar'
 import type { AddressManager } from '@libp2p/interface-address-manager'
 import type { PeerStore } from '@libp2p/interface-peer-store'
 import type { ConnectionManager } from '@libp2p/interface-connection-manager'
 import type { Metrics } from '@libp2p/interface-metrics'
-import type { PeerId } from '@libp2p/interface-peer-id'
 import type { Datastore } from 'interface-datastore'
+import { PeerRouting, peerRouting } from '@libp2p/interface-peer-routing'
+import { PeerDiscovery, peerDiscovery } from '@libp2p/interface-peer-discovery'
```
maschad, line 18: Shouldn't this be imported as `@libp2p/peer-discovery` based on the [interface](https://github.com/libp2p/js-libp2p-interfaces/blob/master/packages/interface-peer-discovery/src/index.ts#LL12C1-L12C1) @achingbrain ?

achingbrain: Ah, no - that’s a typo in the comment. Got to get that TS comment verification stuff in!

PR #473, opened 2023-05-05
feat: switch to server mode automatically when addresses change

achingbrain:
If the node is in client mode, and the `self:peer:update` event is emitted, and the event detail contains publicly routable addresses, switch to server mode.

If the event is emitted and we are in server mode and the event detail contains only private addresses, switch back to client mode.

### DIFF
--- a/src/dual-kad-dht.ts
+++ b/src/dual-kad-dht.ts
@@ -83,25 +83,33 @@ class DHTPeerRouting implements PeerRouting {
   }
 }
 
+// see https://github.com/multiformats/multiaddr/blob/master/protocols.csv
+const P2P_CIRCUIT_CODE = 290
+const DNS4_CODE = 54
+const DNS6_CODE = 55
+const DNSADDR_CODE = 56
+const IP4_CODE = 4
+const IP6_CODE = 41
+
 function multiaddrIsPublic (multiaddr: Multiaddr): boolean {
   const tuples = multiaddr.stringTuples()
 
   // p2p-circuit should not enable server mode
   for (const tuple of tuples) {
-    if (tuple[0] === 290) {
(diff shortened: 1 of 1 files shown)


src/dual-kad-dht.ts
```diff
@@ -81,6 +83,36 @@ class DHTPeerRouting implements PeerRouting {
   }
 }
 
+function multiaddrIsPublic (multiaddr: Multiaddr): boolean {
+  const tuples = multiaddr.stringTuples()
+
+  // p2p-circuit should not enable server mode
+  for (const tuple of tuples) {
```
maschad, line 90: Can we use the [protoCodes](https://github.com/multiformats/js-multiaddr/blob/8582c09da7f601709f843a195286cd6b33a3a201/src/index.ts#L593) function to do these kinds of checks? so something like

``` typescript

const P2P_CIRCUIT_CODE = 290
const DNS4_CODE = 54
// rest of other codes...
function multiaddrIsPublic (multiaddr: Multiaddr): boolean {
	multiaddr.forEach((maddr) => {
		 if (maddr.protoCodes().includes(CIRCUIT_RELAY_CODE)) {
			return false
		 }
		 
		 if (maddr.protoCodes().includes(DNS4_CODE) {
			return true
		}
	})

 	// other checks
 	
 	return false
}

```

maschad (changes requested):
Looks good but left a suggestion

src/dual-kad-dht.ts
achingbrain: Calling `protoCodes()` repeatedly like this is expensive.

I've swapped the magic numbers out for named constants instead, hopefully that's a bit clearer.

PR #478, opened 2023-05-25
fix: use events to delay before self-query

achingbrain:
Instead of debouncing and using timeouts to wait for DHT peers before running the initial self-query, instead get the routing table to emit events when peers are added or removed - if the table is empty when we run the self-query, wait for the `peer:add` event before continuing.

Improves startup time.

Adds tests for the query-self component.

### DIFF
--- a/package.json
+++ b/package.json
@@ -181,6 +181,7 @@
     "it-take": "^3.0.1",
     "multiformats": "^11.0.0",
     "p-defer": "^4.0.0",
+    "p-event": "^5.0.1",
     "p-queue": "^7.3.4",
     "private-ip": "^3.0.0",
     "progress-events": "^1.0.0",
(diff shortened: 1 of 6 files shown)


maschad (commented):
Appreciate this improvement, a more methodical way to run self-queries.

My only question, is this a breaking change with introduction of the [QuerySelfComponents](https://github.com/libp2p/js-libp2p-kad-dht/pull/478/files#diff-4f7402fafb86ee87a9a0a7b6fc0791864a9779af5de180745014a353e40c585dR49) instead of `KadDHTComponents` ?

Otherwise looks good.

achingbrain:
It’s an internal component so not a breaking change.

maschad (approved):

